PR #52, opened 2013-08-19
Export compile_commands.json in build/ so it's easy to work with several vim plugins.

byzhang:


denji:
@byzhang `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON .`

PR #270, opened 2014-09-06
FT-591 fixup valgrind block_allocator_test fails

BohuTANG:
If forward is false, the first bp point to blocks_array[used - 1], so the bp[1] is uninitialised


prohaska7:
a different bug fix was used.

PR #272, opened 2014-11-27
posix_memalign() is allowed to return NULL if the size is zero.  toku_xm...

kuszmaul:
posix_memalign() is allowed to return NULL if the size is zero.  toku_xmalloc_aligned() has an overactive assertion that complains in this situation. The assertion triggers when running ft-test with an implementation that returns NULL on zero-sized values.  This patch allows ft-test run to completion with such an implementation of posix_memalign().


prohaska7:
merged onto master

PR #281, opened 2015-06-17
FT-669 : Refactor status to it's own file/sybsystem within ft-index

georgelorchpercona:
Created new source files ft/ft-status.h and .cc to sort of "C++" the
way ft status values are handled. The problem is that each sub-system
handled stats slightly differently and usually used some static structs
local to only one.cc file. This made updating related stats across
files difficult, needing to add functions to expose stats to other
areas. Also made looking up stats a pain, needing to know were they
lived. Now, all stat enums for ft live in a single header and all stat
declarations are in a single .cc file and it is easy to manipulate/add
stats from anywhere.

### DIFF
--- a/ft/CMakeLists.txt
+++ b/ft/CMakeLists.txt
@@ -36,6 +36,7 @@ set(FT_SOURCES
   ft-flusher
   ft-hot-flusher
   ft-ops
+  ft-status
   ft-test-helpers
   ft-verify
   loader/callbacks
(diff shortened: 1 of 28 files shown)


georgelorchpercona:
jenkins http://jenkins.percona.com/view/TokuDB/job/ft-index-param/10/

georgelorchpercona:
Made minor changes as suggested by Christian, rebuilt and repushed.

PR #282, opened 2015-06-18
FT-670

georgelorchpercona:
Expose 'Engine' status as 'GLOBAL' status for DB Changed all FT related and some environmental related status to appear in both 'engine' and 'global' status lists for easier parsing and usage in MySQL. MX and SE should not be impacted by the change but needs proper verification.

Based on FT-669.

jenkins running : http://jenkins.percona.com/view/TokuDB/job/ft-index-param/11/

### DIFF
--- a/ft/ft-status.cc
+++ b/ft/ft-status.cc
@@ -97,15 +97,15 @@ PATENT RIGHTS GRANT:
 LE_STATUS_S le_status;
 void LE_STATUS_S::init() {
     if (m_initialized) return;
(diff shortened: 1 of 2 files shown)


georgelorchpercona:
Had some merge issue, fixed, new jenkins : http://jenkins.percona.com/view/TokuDB/job/ft-index-param/13

christianrober:
Looks good to me.  Side Note: I don't know why we never added column names for some of the variables.  I assume its an improvement to legibility, if nothing else.  I like the new code formatting too, thanks George.

PR #284, opened 2015-07-02
don't call malloc_usable_size() so much

kuszmaul:
In some situations, calls to malloc_usable_size() were taking significant CPU cycles.  (Vadim measured 8% on a workload with a fanout of 128.)  We can remember the usable size of the message buffers, saving those CPU cycles.

(There is a gratuituous commit mentioned in this pull request: f59ea95 was already pulled as part of a previous request.  I don't know how to get rid of it; maybe I shouldn't even try.


georgelorchpercona:
Superseded by https://github.com/Tokutek/ft-index/pull/288. Will merge 288 once regression testing has completed. Thanks much Bradley for the contribution.

PR #290, opened 2015-07-11
Unlocked signal evictor

prohaska7:
helgrind and drd complain when the evictor cond var is signaled with the associated mutex held.  this change grabs the evictor mutex when signaling the evictor.  an alternate design could suppress the error.  IMO, the code change is safer.


georgelorchpercona:
https://tokutek.atlassian.net/browse/FT-665

PR #292, opened 2015-07-11
Circular buffer assert

prohaska7:
the circular buffer test had a race problem in it that occurs 1 of 200 test runs.  i fixed the test.  since the circular buffer is not used by the fractal tree software, i deleted all of the circular buffer code.


georgelorchpercona:
https://tokutek.atlassian.net/browse/FT-663

PR #294, opened 2015-07-24
FT-685 : introduce branch prediction macros

georgelorchpercona:
Added simple FT_LIKELY and FT_UNLIKELY branch prediction macros similar to InnoDB UNIV_LIKELY and UNIV_UNLIKELY.
Functional only in GCC environments otherwise they are noops and simply evaluate the expression as if they didn't exist.

Macros not applied to any branches yet, just putting in place for use in ongoing work.

### DIFF
--- a/portability/toku_portability.h
+++ b/portability/toku_portability.h
@@ -331,3 +331,28 @@ void toku_portability_destroy(void);
 static inline uint64_t roundup_to_multiple(uint64_t alignment, uint64_t v) {
     return (v + alignment - 1) & ~(alignment - 1);
 }
+
+// Branch prediction macros.
+// If supported by the compiler, will hint in inctruction caching for likely
+// branching. Should only be used where there is a very good idea of the correct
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
What about already-existing

#define toku_likely(x)   __builtin_expect((x),1)
#define toku_unlikely(x) __builtin_expect((x),0)

in the same file?

georgelorchpercona:
Where do you see that Laurynas? I grep nor see toku_likely or toku_unlikely anywhere in ft-index. I do see a few places in ft-index that are using the __builtin_expect directly though:

./ft/comparator.h:169:            if (__builtin_expect(toku_dbt_is_infinite(a) || toku_dbt_is_infinite(b), 0)) {
./ft/comparator.h:175:                       && __builtin_expect(dbt_has_memcmp_magic(b), 1)) {
./ft/ft-ops.cc:4563:    if (__builtin_expect(c != 0, 1)) {

laurynas-biveinis:
Ah, I was looking at a different branch than trunk, sorry.

PR #297, opened 2015-07-30
FT-689 : Add option to allow disabling of partial eviction of nodes

georgelorchpercona:
Requested by Vadim T. - "For start I propose we create an option --tokudb-enable-partial-eviction (ON by default)."

Jenkins having issues now so no jenkins matrix test. local ctest passed as expected.


ft/cachetable/cachetable.cc
```diff
@@ -4282,6 +4287,14 @@ void evictor::fill_engine_status() {
     CT_STATUS_VAL(CT_LONG_WAIT_PRESSURE_TIME) = read_partitioned_counter(m_long_wait_pressure_time);
 }
 
+void evictor::set_enable_partial_eviction(bool enabled) {
+    m_enable_partial_eviction = enabled;
+}
+
+bool evictor::get_enable_partial_eviction(void) {
```
christianrober: I would consider making this method const.

christianrober:
Looks good to me.

ft/cachetable/cachetable.cc
georgelorchpercona: Yes, I already have done this per our Slack conversation. I haven't pushed it yet as I'm still working on introducing a test case that validates disabling of partial evictions. It would also be a good idea to test compress_buffers_before_eviction since there is no test for it that I could find and it is not documented anywhere.

georgelorchpercona:
Alrighty, made const and modified a test case to validate the new functionality. ctest passes locally.

PR #299, opened 2015-07-31
FT-678 : add thread pool size controls for CLIENT, CACHETABLE or CHECKPOINT

georgelorchpercona:
Must be specified at ENV.open, not dynamic.

### DIFF
--- a/ft/cachetable/cachetable.h
+++ b/ft/cachetable/cachetable.h
@@ -108,7 +108,14 @@ uint32_t toku_get_cleaner_iterations_unlocked (CACHETABLE ct);
 // create and initialize a cache table
 // size_limit is the upper limit on the size of the size of the values in the table
 // pass 0 if you want the default
-int toku_cachetable_create(CACHETABLE *result, long size_limit, LSN initial_lsn, struct tokulogger *logger);
+int toku_cachetable_create_ex(CACHETABLE *result, long size_limit,
+                           unsigned long client_pool_threads,
+                           unsigned long cachetable_pool_threads,
+                           unsigned long checkpoint_pool_threads,
+                           LSN initial_lsn, struct tokulogger *logger);
+
+#define toku_cachetable_create(r, s, l, o) \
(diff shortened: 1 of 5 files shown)


ft/cachetable/cachetable.h
```diff
@@ -108,7 +108,14 @@ uint32_t toku_get_cleaner_iterations_unlocked (CACHETABLE ct);
...
 // pass 0 if you want the default
-int toku_cachetable_create(CACHETABLE *result, long size_limit, LSN initial_lsn, struct tokulogger *logger);
+int toku_cachetable_create2(CACHETABLE *result, long size_limit,
+                           unsigned long client_pool_threads,
+                           unsigned long cachetable_pool_threads,
+                           unsigned long checkpoint_pool_threads,
+                           LSN initial_lsn, struct tokulogger *logger);
+
+#define toku_cachetable_create(r, s, l, o) \
+    toku_cachetable_create2(r, s, 0, 0, 0, l, 0);
```
christianrober: Instead of adding a MACRO, and using a less-than-optimal name with a '2' appended, couldn't we just put these arguments at the end of the function declaration and give them default values?

The other option is to create a second function with a different name that also creates a cacheable but with the optional manual thread count parameters.  Something like: toku_create_cachetable_with_thread_counts().  Then people would know there are two different behaviors and be forced to make an explicit choice while coding.

georgelorchpercona: Yes, I started out that way, well, without the default values and found that hundreds of test cases needed changing. I'm not particularly fond of mixing C++ default values with straight C type API calls, but I do get your point. This was an old habit of sorts from InnoDB and others.

In reality I would much rather a multi-phased creation-initialization-start approach that would be something like:
`toku_cachetable_create(...);
toku_cachetable_set_something(...);
toku_cachetable_set_something(...);
toku_cachetable_set_something(...);
toku_cachetable_start(...);`

PR #301, opened 2015-08-04
DB-818 fix reserved filenum overflow

BohuTANG:
[summary]
We have got an error on our production:
toku_cachetable_openfd_with_filenum: Assertion 'filenum.fileid != FILENUM_NONE.fileid' failed

From the gdb statck, I got:
(gdb) p reserved_filenum
 = {fileid = 4294967295}

Deep in the 'reserve_filenum' codes, found that:
a) this function just return the m_next_filenum_to_use, not the un-used filenum.
b) m_next_filenum_to_use is UINT32， if we have many ft opens, it maybe overflow

From the crash instance data dir, we got one database has 180,000 tokudb files, so it's easy to trigger the assert.

Casey Brown has reported this issue here:
https://tokutek.atlassian.net/browse/DB-818

[testcase]
ft/tests/cachetable-reserve-filenum.cc

Reviewed By: Rich

### DIFF
--- a/ft/cachetable/cachetable.cc
+++ b/ft/cachetable/cachetable.cc
@@ -4767,6 +4767,7 @@ void cachefile_list::remove_stale_cf_unlocked(CACHEFILE cf) {
 
 FILENUM cachefile_list::reserve_filenum() {
     // taking a write lock because we are modifying next_filenum_to_use
+    FILENUM filenum = FILENUM_NONE;
     write_lock();
     while (1) {
         int r = m_active_filenum.find_zero<FILENUM, cachefile_find_by_filenum>(m_next_filenum_to_use, nullptr, nullptr);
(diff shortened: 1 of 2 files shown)


georgelorchpercona:
Thanks for the contribution and the recent fix. I need two more things before I can merge this fix:
1) In the cachetable-reserve-filenum.cc test file, please change the license text to be the new PerconaFT license text that you can now find in any of the other source files.
2) Please review the new contribution licensing requirements in https://github.com/percona/PerconaFT/blob/master/README.md "Contributing" section and provide the proper license text as part of the commit and pull request comments.

Thank you again for you efforts and contribution to the project.

BohuTANG:
Change to https://github.com/percona/PerconaFT/pull/302

PR #304, opened 2015-08-13
FT-691 (CMake fails on case-insensitive FS, such as Mac OS X)

laurynas-biveinis:
Swap option(USE_ETAGS ... ) and option(USE_CTAGS ... ) in
cmake_modules/TokuBuildTagDatabases.cmake, so that if USE_ETAGS is
processed first and USE_CTAGS then is properly disabled on Mac OS X.


laurynas-biveinis:
With this patch -DUSE_CTAGS=OFF -DUSE_ETAGS=ON, -DUSE_CTAGS=ON -DUSE_ETAGS=OFF, -DUSE_CTAGS=OFF -DUSE_ETAGS=OFF, and specifying nothing allows configure to complete and outputs the right tag-building rules.

PR #309, opened 2015-08-18
Introduce TOKU_FILE struct as a wrapper for FILE

laurynas-biveinis:
This is a prerequisite for adding performance schema and possibly
other kinds of instrumentation.

Create a struct TOKU_FILE that at the moment contains only a FILE *
field. Update functions to pass TOKU_FILE instead of FILE around.

If desired, it is possible to convert TOKU_FILE to a C++ class and
make its FILE field private in the future.

A new header file is created for this struct, toku_instrumentation.h.
This file is going to be used in the later instrumentation-introducing
commits.

The branch contains FT-694 fix merged as a prerequisite. Tested by verifying that ctest does not introduce any regressions against FT-694 branch (but see FT-695)

### DIFF
--- a/portability/file.cc
+++ b/portability/file.cc
@@ -277,25 +277,29 @@ toku_os_pwrite (int fd, const void *buf, size_t len, toku_off_t off) {
     return result;
 }
 
-FILE * 
+TOKU_FILE *
 toku_os_fdopen(int fildes, const char *mode) {
     FILE * rval;
     if (t_fdopen)
 	rval = t_fdopen(fildes, mode);
     else 
 	rval = fdopen(fildes, mode);
-    return rval;
+    if (FT_UNLIKELY(rval == nullptr))
+        return nullptr;
+    return new TOKU_FILE(rval);
 }
     
 
-FILE *
+TOKU_FILE *
 toku_os_fopen(const char *filename, const char *mode){
     FILE * rval;
     if (t_fopen)
 	rval = t_fopen(filename, mode);
     else
 	rval = fopen(filename, mode);
-    return rval;
+    if (FT_UNLIKELY(rval == nullptr))
+        return nullptr;
+    return new TOKU_FILE(rval);
 }
 
 int 
@@ -328,17 +332,18 @@ toku_os_open_direct(const char *path, int oflag, int mode) {
 }
 
 int
-toku_os_fclose(FILE * stream) {  
+toku_os_fclose(TOKU_FILE * stream) {
     int rval = -1;
     if (t_fclose)
-	rval = t_fclose(stream);
+	rval = t_fclose(stream->file);
     else {                      // if EINTR, retry until success
 	while (rval != 0) {
-	    rval = fclose(stream);
+	    rval = fclose(stream->file);
 	    if (rval && (errno != EINTR))
 		break;
 	}
     }
+    delete stream;
     return rval;
 }
 
--- a/portability/toku_instrumentation.h
+++ b/portability/toku_instrumentation.h
@@ -1,0 +1,47 @@
+/* -*- mode: C++; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+// vim: ft=cpp:expandtab:ts=8:sw=4:softtabstop=4:
+#ident "$Id$"
+/*======
+  This file is part of PerconaFT.
+
+
+  Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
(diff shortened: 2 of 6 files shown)


portability/toku_instrumentation.h
```diff
@@ -0,0 +1,47 @@
...
+  along with PerconaFT.  If not, see <http://www.gnu.org/licenses/>.
+  ======= */
+
+#ident "Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved."
+
+#pragma once
+
+#include <stdio.h> // FILE
+
+struct TOKU_FILE
```
christianrober: I was hoping that the file containing the TOKU_FILE struct/class would be in a similarly named file.  One reason:  consumers of this object should not know anything about whether it is instrumented or not.  These consumers would just have a dependency on something like toku_file.h.  This is especially true if we add methods to the object and it needs a corresponding .cpp/.cc file for the implementation.

That said...
- I understand that the design of the instrumentation will be optionally compiled away, so dependents can't access instrumentation details without compiler errors when the fields are not there. 
- The instrumentation for a file needs to live somewhere, and this is as a logical place as any.  
- I also recognize the possible redundancy of having a file named toku_file.h and then a separate file header with the compiled in instrumentation.

portability/file.cc
```diff
@@ -277,25 +277,29 @@ toku_os_pwrite (int fd, const void *buf, size_t len, toku_off_t off) {
...
+TOKU_FILE *
 toku_os_fdopen(int fildes, const char *mode) {
     FILE * rval;
     if (t_fdopen)
 	rval = t_fdopen(fildes, mode);
     else 
 	rval = fdopen(fildes, mode);
-    return rval;
+    if (FT_UNLIKELY(rval == nullptr))
+        return nullptr;
```
georgelorchpercona: Note that this function will return nullptr on failure, not a TOKU_FILE with the FILE* component NULL...

portability/file.cc
```diff
@@ -277,25 +277,29 @@ toku_os_pwrite (int fd, const void *buf, size_t len, toku_off_t off) {
...
+TOKU_FILE *
 toku_os_fopen(const char *filename, const char *mode){
     FILE * rval;
     if (t_fopen)
 	rval = t_fopen(filename, mode);
     else
 	rval = fopen(filename, mode);
-    return rval;
+    if (FT_UNLIKELY(rval == nullptr))
+        return nullptr;
```
georgelorchpercona: Note that this function will return nullptr on failure, not a TOKU_FILE with the FILE* component NULL...

georgelorchpercona:
There are several places in this new code where (TOKU_FILE*)something->file is being tested. Shouldn't TOKU_FILE be an opaque type? If the toku_*open functions fail, they will return a (TOKU_FILE*)nullptr, not a valid TOKU_FIE with the FILE* component as a nullptr. So there are many tests that test "if (<something>->file) {...}" that should be either "if (<something>) {...}" or perhaps add a new validation function to toku_portability like "bool toku_os_is_valid_file(const TOKU_FILE* fd) { return fd && fd->file ? true : false; }" and call that from anywhere that needs to test the validity of TOKU_FILE.

georgelorchpercona:
Also it would be great to have an FT - JIRA task to associate this with even though this doesn't have a real user facing component at the moment.

portability/toku_instrumentation.h
laurynas-biveinis: Proper header file organisation is still something we are trying to figure out. The current prototype has the following scheme: toku_portability.h includes toku_instrumentation.h (declaring instrumentation stub structs and functions), which optionally includes toku_mysql.h (declaring same structs and functions for MySQL PFS).

TOKU_FILE falls into toku_instrumentation.h by the path of least resistance then. But the above scheme has things declared very much out of place (mutex stuff etc) to resolve cyclic dependencies.

The above means that the current header scheme is not final. I can try introducing toku_file.h and move all the file-related declarations there too at the same time.

portability/file.cc
laurynas-biveinis: Ack, see below

laurynas-biveinis:
TOKU_FILE should become an opaque type, and the PR description mentions something towards that direction, but it's outside the scope of current PR. I will make failing function to return a TOKU_FILE with a nullptr FILE instead of nullptr TOKU_FILE. This seems to be more conductive towards future opaque type (ie. file->is_valid() or similar)

PR #307, opened 2015-08-17
document the problems with the helgrind annotations of unsafe fetch and set

prohaska7:
document the problems with the helgrind annotations of unsafe fetch and set.  see http://prohaska7.blogspot.com/2015/08/races-in-tokuft-race-detector.html

    Copyright (c) 2015, Rich Prohaska
    All rights reserved.

    Redistribution and use in source and binary forms, with or without modification, are permitted provided that the following conditions are met:

    1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

    2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribu

    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS

PR #310, opened 2015-08-18
fix memory leak in bnc insert test found with address sanitizer

prohaska7:
fix memory leak found with address sanitizer.  for some reason, ctest -DExperimentalMemCheck does not run the bnc-insert test with valgrind.   using address sanitizer with clang-3.5 on ubuntu 14.04.
    
    All rights reserved.
    
    2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribu
    
    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS

### DIFF
--- a/ft/txn/rollback.cc
+++ b/ft/txn/rollback.cc
@@ -92,7 +92,7 @@ static inline PAIR_ATTR make_rollback_pair_attr(long size) {
 PAIR_ATTR
 rollback_memory_size(ROLLBACK_LOG_NODE log) {
     size_t size = sizeof(*log);
-    if (&log->rollentry_arena) {
+    if (1) { // if (&log->rollentry_arena) {
         size += log->rollentry_arena.total_footprint();
     }
     return make_rollback_pair_attr(size);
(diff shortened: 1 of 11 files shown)


ft/txn/rollback.cc
```diff
@@ -92,7 +92,7 @@ static inline PAIR_ATTR make_rollback_pair_attr(long size) {
 PAIR_ATTR
 rollback_memory_size(ROLLBACK_LOG_NODE log) {
     size_t size = sizeof(*log);
-    if (&log->rollentry_arena) {
+    if (1) { // if (&log->rollentry_arena) {
```
christianrober: Do we want to remove this predicate check?  I assume the rollentry_arena is always non-null and the predicate check can be removed completely.

...Or maybe this change is related to some other work and snuck in by accident?

prohaska: That line should be deleted.

georgelorchpercona: Looks like it was introduced here https://github.com/percona/PerconaFT/pull/307, so both need fixing.

prohaska7:
redo

PR #312, opened 2015-08-25
added profiling option to allow profiling/debug. Jira ticket BLD-300

abychko:
[+] new option for cmake (-DPROFILING)
[+] added -fno-omit-frame-pointer by request from Vadim and turned it on only if -DPROFILING is set

tested 3 times to make sure it works:
1) without -DPROFILING. result - -fno-omit-frame-pointer is not set
2) with -DPROFILING=on. result - -fno-omit-frame-pointer is set
3) with -DPROFILING=off. result - -fno-omit-frame-pointer is not set

### DIFF
--- a/cmake_modules/TokuSetupCompiler.cmake
+++ b/cmake_modules/TokuSetupCompiler.cmake
@@ -117,7 +117,6 @@ if (NOT CMAKE_CXX_COMPILER_ID MATCHES Clang)
     )
 endif ()
 
-## 
 option (PROFILING "Allow profiling and debug" OFF)
 if (PROFILING)
   set_cflags_if_supported(
--- a/cmake_modules/TokuSetupCompiler.cmake
+++ b/cmake_modules/TokuSetupCompiler.cmake
@@ -117,6 +117,14 @@ if (NOT CMAKE_CXX_COMPILER_ID MATCHES Clang)
     )
 endif ()
 
+## 
+option (PROFILING "Allow profiling and debug" OFF)
+if (PROFILING)
+  set_cflags_if_supported(
+    -fno-omit-frame-pointer
+  )
+endif ()
+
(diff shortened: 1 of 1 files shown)


cmake_modules/TokuSetupCompiler.cmake
```diff
@@ -117,6 +117,14 @@ if (NOT CMAKE_CXX_COMPILER_ID MATCHES Clang)
     )
 endif ()
 
+##
```
christianrober: I would consider removing this empty comment.

cmake_modules/TokuSetupCompiler.cmake
```diff
@@ -117,6 +117,13 @@ if (NOT CMAKE_CXX_COMPILER_ID MATCHES Clang)
     )
 endif ()
 
+option (PROFILING "Allow profiling and debug" OFF)
+if (PROFILING)
+  set_cflags_if_supported(
```
vadimtk: what about cxxflags ?

abychko:
it is not needed, see macros:

 61 ## adds a compiler flag if the compiler supports it
 62 macro(set_cflags_if_supported)
 63   foreach(flag ${ARGN})
 64     check_c_compiler_flag(${flag} HAVE_C_${flag})
 65     if (HAVE_C_${flag})
 66       set(CMAKE_C_FLAGS "${flag} ${CMAKE_C_FLAGS}")
 67     endif ()
 68     check_cxx_compiler_flag(${flag} HAVE_CXX_${flag}) <<<<<<
 69     if (HAVE_CXX_${flag})
 70       set(CMAKE_CXX_FLAGS "${flag} ${CMAKE_CXX_FLAGS}") <<<<<
 71     endif ()
 72   endforeach(flag)
 73 endmacro(set_cflags_if_supported)

PR #315, opened 2015-08-25
    document the problems with the helgrind annotations of unsafe fetch and set .    generalize the unsafe fetch and set function names for helgrind, drd, and tsan.

prohaska7:
document the problems with the helgrind annotations of unsafe fetch and set
    generalize the unsafe fetch and set function names for helgrind, drd, and tsan.
    
    All rights reserved.
    
    2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the dist
    
    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITN

### DIFF
--- a/buildheader/make_tdb.cc
+++ b/buildheader/make_tdb.cc
@@ -430,7 +430,9 @@ static void print_db_key_range_struct (void) {
 
 static void print_db_lsn_struct (void) {
     field_counter=0;
-    sort_and_dump_fields("db_lsn", false, NULL);
+    /* A dummy field to make sizeof(DB_LSN) equal in C and C++ */
+    const char *extra[] = { "char dummy", NULL };
+    sort_and_dump_fields("db_lsn", false, extra);
 }
 
 static void print_dbt_struct (void) {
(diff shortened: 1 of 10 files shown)


christianrober:
Looks good to me.  The comments are very helpful, thanks Rich.  I could use some clarification though:
 
  I would expect that the code `if (0) TOKU_VALGRIND_HG_DISABLE_CHECKING` was left there to be a reminder to re-enable in the future.  I assume that this future date is when Helgrind fixes its support of those annotations.  In the meantime I am surprised we did not need to disable more than one test.  Should we expect to see other Helgrind failures which amount to a false failure because of the missing annotations?

prohaska7:
I left the helgrind enable/disable code in hoping that helgrind will work someday.  the problem is described here: http://prohaska7.blogspot.com/2015/08/races-in-tokuft-race-detector.html

there are some tests that use helgrind and do not depend on enable/disable working AFAIK, so those are left in.  otherwise, drd is used.

georgelorchpercona:
https://tokutek.atlassian.net/browse/FT-698

PR #319, opened 2015-09-09
FT-699 (Missing field initialisers failing build)

laurynas-biveinis:
Add missing (and in case of mutexes and cond structs, replace with the
correct ones) initializers to fix clang -Wmissing-field-initializer
build-failing warnings.


christianrober:
Looks good to me.

laurynas-biveinis:
Tested that builds on Mac, in Percona Server and standalone, and on Linux, standalone, debug and release configs on Ubuntu 14.10

PR #326, opened 2015-09-26
Returned null pointer checks in DMT and turned off problematic clang flags.

christianrober:
The recent commit to remove null checks and make the clang warnings (as errors) go away needs to be reverted. It has broken some tests, including the main DMT test.

Background: The production code instantiates the DMT with pointer types that are never null, which triggers the warning about the justifiably unnecessary checks.  However, the DMT test DOES instantiate the template with NULL-able arguments.  Presumably Clang doesn't make an exception for templates whose instantiation is potentially ambiguous with respect to null-able types...

So removing the checks from the code removes the warnings but breaks the test.  So, instead, we just disable those two warnings for clang only.  If this issue ever gets resolved we can re-enable them.

The other option is to remove all the null-able arguments from the associated tests, but that requires refactoring of various paths within the DMT template, which causes an arguably unnecessary amount of churn for suppressing a compiler warning.... Hence this solution.

### DIFF
--- a/cmake_modules/TokuSetupCompiler.cmake
+++ b/cmake_modules/TokuSetupCompiler.cmake
@@ -188,6 +188,16 @@ if (NOT CMAKE_CXX_COMPILER_ID STREQUAL Clang)
   set_cflags_if_supported(-Wcast-align)
 endif ()
 
+## Clang's pointer comparison warnings create false
+## negatives when instantiating templates that are
+## sometimes instantiated with null pointers and
+## other times with non-null local variable addresses.
+## The DMT implementation and test are examples.
+if (CMAKE_CXX_COMPILER_ID STREQUAL Clang)
+   set_cflags_if_supported(-Wno-tautological-pointer-compare)
+   set_cflags_if_supported(-Wno-pointer-bool-conversion)
+endif()
+
 ## always want these
 set(CMAKE_C_FLAGS "-Wall -Werror ${CMAKE_C_FLAGS}")
 set(CMAKE_CXX_FLAGS "-Wall -Werror ${CMAKE_CXX_FLAGS}")
--- a/util/dmt.cc
+++ b/util/dmt.cc
@@ -870,26 +870,42 @@ void dmt<dmtdata_t, dmtdataout_t, dmtwriter_t>::rebalance(subtree *const subtree
 
 template<typename dmtdata_t, typename dmtdataout_t, typename dmtwriter_t>
 void dmt<dmtdata_t, dmtdataout_t, dmtwriter_t>::copyout(uint32_t *const outlen, dmtdata_t *const out, const dmt_node *const n) {
-    *outlen = n->value_length;
-    *out = n->value;
+    if (outlen) {
+        *outlen = n->value_length;
+    }
+    if (out) {
+        *out = n->value;
+    }
 }
 
 template<typename dmtdata_t, typename dmtdataout_t, typename dmtwriter_t>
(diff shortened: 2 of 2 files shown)


georgelorchpercona:
I am very close to automatically rejecting any OSX/clang specific pull requests. It seems that 50% or more of them lately are breaking supported platform builds and the contributors are not performing proper  due-diligence to prevent such issues.

christianrober:
Yes, it definitely seems that contributors, in general, don't run the changes through the test suite on Mac OS X/Clang as rigorously as they do for Linux/gcc.

Do we want to just revert the last merge instead of applying this patch?

georgelorchpercona:
For this case, yes, I think it is probably the easiest thing.

On 09/25/2015 09:52 PM, Christian wrote:
>
> Yes, it definitely seems that contributors, in general, don't run the 
> changes through the test suite on Mac OS X/Clang as rigorously as they 
> do for Linux/gcc.
>
> Do we want to just revert the last merge instead of applying this patch?
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/percona/PerconaFT/pull/326#issuecomment-143398973>.
>

-- 
George O. Lorch III
Software Engineer, Percona
US/Arizona (GMT -7)
skype: george.ormond.lorch.iii

prohaska7:
The problem still exists with clang 3.8 on Linux.  Either the non-null
attribute us wrong or the test is wrong.

christianrober:
Well, it is more than just the test that needs to be changed.  The DMT class does pass null arguments, during `insert`.  Here is a diff of a fix I concocted for the both the test failures and a change to the DMT insertion logic:
https://github.com/christianrober/PerconaFT/commit/bb26df1ea83a54a7caeb54ddb8e8c4843dbb2c2a

So, yes, in order to support building on clang, regardless of platform, we either need to disable those two warnings or make alterations similar to the ones above.

On the one hand, I think instantiated templates introduce an ambiguity across the compilation stage that Clang cannot resolve.  Arguably, Clang should not report the warning for instantiated templates because they could be validly instantiated elsewhere, as is the case for the DMT code.

On the other hand one can assume that the DMT null argument behavior is an optimization and could be changed.  I am assuming this logic is an optimization to avoid unnecessary dereferences and the copying of fields/pointers, while keeping the top level API consistent with respect to argument count.

While this does smell like an unnecessary and premature optimization, there may be a valid scenario for it that I have yet to discern from the code.  Assuming for a moment it isn't necessary, a valid third option would be to keep the warnings, keep the removal of the pointer value checks, then change the insertion code path, and then update the tests:

1) We could create a new overload of find_zero that only takes two arguments and doesn't (can't) copy-out the size and data variables.

2) Update the current test cases as shown above.  These test fixes don't seem to have any corollary in the production code, so they should be safe to update or remove.  Again, the intention of these test cases is not clear to me, but the mechanics are simple enough.  Perhaps adding more overloads to the DMT API, based on the test cases, is the best way of resolving the test state.

laurynas-biveinis:
The combination of __attribute__((nonnull)) and if (foo == nullptr) is a bomb waiting to explode. The attribute annotation is useless and dangerous at the same time. The only thing for diagnostics it does is to issue a warning in the case of "foo(NULL)" calls. GCC does not even warn for "void *bar = NULL; foo(bar);" case, limiting the value of diagnostics. The other thing it does is to enable silent compilation away any null checks for args inside the annotated functions, because hey, cannot happen! Which is the dangerous part (and identical to the original PR). I'd suggest to remove any such annotations globally in addition to any other fixes here.

laurynas-biveinis:
Filed my last comment as https://tokutek.atlassian.net/browse/FT-704 per @christianrober's request

PR #328, opened 2015-10-01
FT-703 - 'pool_name' status missing actual pool name in engine status

georgelorchpercona:
- Added pool names to string descriptions for engine status values.


christianrober:
Looks good to me.

PR #330, opened 2015-10-02
  [FT-681] Transactional inconsistency with rollback

Jun-Yuan:
Remake the pull request
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/36/

### DIFF
--- a/src/tests/rollback-inconsistency.cc
+++ b/src/tests/rollback-inconsistency.cc
@@ -1,0 +1,187 @@
+/* -*- mode: C++; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+// vim: ft=cpp:expandtab:ts=8:sw=4:softtabstop=4:
+#ident "$Id$"
+/*======
+This file is part of PerconaFT.
+
+
+Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
+
+    PerconaFT is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License, version 2,
+    as published by the Free Software Foundation.
+
+    PerconaFT is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with PerconaFT.  If not, see <http://www.gnu.org/licenses/>.
+
+----------------------------------------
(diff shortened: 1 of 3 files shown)


src/tests/rollback-inconsistency.cc
```diff
@@ -0,0 +1,187 @@
...
+        DBT val = { .data = str, .size = 220 };
+        r = db->put(db, txn, &key, &val, 0);
+        assert_zero(r);
+    }
+
+    r = txn->commit(txn, 0); assert_zero(r);
+    
+    r = parent->abort(parent); assert_zero(r);
+    //try to search anything in the lost range
+        int k = htonl(200);
```
christianrober: Why is this still indented?  Did you intend to perform more than one `db->get()` in a loop?

Jun-Yuan: Yeah. You are too smart. I was doing a loop before then I thought it is enough to do one quick lookup for proof-of-concept. Then I forgot I already deleted the loop...:(.............will remove the indent later today...in the middle of chasing read-commit.

src/tests/rollback-inconsistency.cc
```diff
@@ -0,0 +1,187 @@
...
+
+    r = txn->commit(txn, 0); assert_zero(r);
+    //again eviction due to the memory pressure. 256 rows is the point when that rollback log spills out. The spilled node
+    //will be written back but will not be dirtied by including rollback nodes from child txn(in which case the bug is bypassed).
+ 
+    r = env->txn_begin(env, parent, &txn, 0); assert_zero(r);
+
+    // populate
+    for (int i = 256; i < 512; i++) {
+        int k = htonl(i);
```
christianrober: The contents of this "populate" loop look identical to the previous two loops, aside from the initial and ending conditions of `i`.  Perhaps they should be refactored into a function called `populate_in_range` or `put_keys_with_child_txn` that takes the parent TXN, the db and the two `i` bounds.  Then this function would be called three times, wrapped by the creation and aborting of the parent transaction, followed by the `get` step to verify the bug.  That would be a lot easier to read/understand and avoid code duplication:

```
DB_TXN *parent = NULL;
int r = env->txn_begin(env, NULL, &parent, 0); assert_zero(r);
populate(env, db, parent, 0, 128);
populate(env, db, parent, 128, 256);
populate(env, db, parent, 256, 512);
r = parent->abort(parent); assert_zero(r);
...
```

Jun-Yuan:
OK. That sounds good. I will modify as the suggestion. Thank you.

PR #332, opened 2015-10-06
 [FT-681] Transactional inconsistency with rollback

Jun-Yuan:
The indent and refactoring of funcs. re-commited and re-request the pull
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/36/

### DIFF
--- a/src/tests/rollback-inconsistency.cc
+++ b/src/tests/rollback-inconsistency.cc
@@ -1,0 +1,161 @@
+/* -*- mode: C++; c-basic-offset: 4; indent-tabs-mode: nil -*- */
+// vim: ft=cpp:expandtab:ts=8:sw=4:softtabstop=4:
+#ident "$Id$"
+/*======
+This file is part of PerconaFT.
+
+
+Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
+
+    PerconaFT is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License, version 2,
+    as published by the Free Software Foundation.
+
+    PerconaFT is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with PerconaFT.  If not, see <http://www.gnu.org/licenses/>.
+
+----------------------------------------
+
+    PerconaFT is free software: you can redistribute it and/or modify
+    it under the terms of the GNU Affero General Public License, version 3,
+    as published by the Free Software Foundation.
+
+    PerconaFT is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Affero General Public License for more details.
+
+    You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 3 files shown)


christianrober:
Looks good to me.

georgelorchpercona:
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/36/

Branch was renamed since jenkins, was FT-691 when it should have been FT-681. Commit is different due to some code formatting difffs from jenkins run and additional comments added to test case.

src/tests/rollback-inconsistency.cc
```diff
@@ -0,0 +1,161 @@
...
+populate_table(int start, int end, DB_TXN * parent, DB_ENV * env, DB * db) {
+    DB_TXN *txn = NULL;
+    int r = env->txn_begin(env, parent, &txn, 0); assert_zero(r);
+    for (int i = start; i < end; i++) {
+        int k = htonl(i);
+        char kk[4]; 
+        char str[220];
+        memset(kk, 0, sizeof kk);
+        memcpy(kk, &k, sizeof k);
+        memset(str,'a', sizeof str-1);
```
BohuTANG: should memset(str, 0, sizeof str) first, or we will get uninitialized error.

kuszmaul: This code is kind of unclean.  Better would be

 int k = htonl(i);
 char str[220];
 memset(str, 'a', sizeof str);
 DBT key = { .data = k, .size = sizeof k};
 DBT val = { .data = str, .size = sizeof str};

1) Don't create a separate character array when k is just fine.  (You might have to cast k to an appropriate pointer when initialing the key dbt).
 2) Don't repeat the literal value "220" in the initializion of val.
 3) Don't memset str twice.  Just set the last entry to 0 if you want a 0-terminated string.  It doesn't look like this test requires a 0-terminates string, however.  In fact, the test doesn't really rely on the value of the string being any particular value. So unless valgrind is complaining, the test is arguably OK with uninitialized str[219].  Valgrind will likely complain, however, when the database is written to disk.

Jun-Yuan: Hi, Thanks @BohuTANG and @kuszmaul for pointing out problems. I actually copied the part of key and value from other code ... I should have been more careful with it.  Will correct these issue together with my next pull request for fixing read-commit bug. Thanks.

kuszmaul: Yes, some of those tests (especially the old ones) were messy...

-Bradley

BohuTANG:
So we should change valgrind suites to thread/addresssanitizer, it's faster and for all tests.

prohaska7:
I would add address sanitizer and thread sanitizer to tests rather than
replace valgrind/helgrind/drd with address/thread sanitizer.  There is
still innovation in the valgrind toolset that we probably want to work with.

The use of the thread sanitizer with the Tokutek code is a bit of a project
since there are lots of deliberately racy variables that are marked with
helgrind annotations so that data races on them are suppressed.

I did an experiment running the tokubackup tests with the thread
sanitizer.  Results are here:
http://prohaska7.blogspot.com/2015/10/tokubackup-versus-sanitizers.html.
 gcov found code uncovered by tests.  Address sanitizer found a memory
leak.  Thread sanitizer found a real problem in tokubackup.

Since the FT software is much larger than tokubackup, converting it to run
with the thread sanitizer will be a larger project.  The main problem is
that helgrind suppressions of memory ranges at run time are not supported
by the thread sanitizer.  So, the method that FT used to annotate racy
variables needs to be replaced.  IMO, all racy variables should use either
be c++ atomic variables OR use explicit functions to load/store so that
thread sanitizer suppressions can be added.

PR #333, opened 2015-10-08
FT-684 : estimated # of rows in a table could become inaccurate after…

georgelorchpercona:
… deletes

* Implemented 'logical row count' for new tables or recounted tables.
* This will not replace the existing physical counting mechanism internally.
* Adds a new, persisted counter value to the ft_header and return this new value in toku_ft_stat64 instead of the current value.
* Counting:
  - New value will increment and decrement at high level calls from db_put, db_delete, etc:
    + toku_ft_insert_unique
    + toku_ft_maybe_insert
    + toku_ft_maybe_delete
  - Lower level leaf entry calls will also return delta values up to a point where they can be applied to the counter and are a result of message application to a leaf node.
    + inject_msg_in_locked_node
  - These deltas are needed for cases where the original count adjustment is not correct for the action:
    + An insert that turns into an update due to a duplicate key. (upsert)
    + A delete where the key doesn’t exist. (blind delete)
    + Transaction rollback. (unapplied rollback messages)
    *NOTE* If any of these three situations occur, the count will still report as somewhat inaccurate until the messages have been applied to all leaf entries affected. This is strictly a message delivery issue.
* The new counter will need some starting state:
  - On a new index, the count will be set to 0 and counted correctly forward from there.
  - On an existing/upgraded index:
    + The count value will be set to some marker value (-1). When in this state, no logical row counting will be applied to this value and the row count estimate will return the current physical row count.
    + A new function will be exposed added that will perform a logical recount of all rows. Upon successful completion the logical row count in the ft_header will be set to this recounted value.
    + int recount_rows(DB*, int (*progress_callback)(uint64_t count, uint64_t deleted)void* progress_extra), void* progress_extra);
    + Algorithm will recount all real, logical rows in the tree and update the logical count value in ft_header.
    + progress_callback will be passed the current key count and the number of deleted ules seen since the last call.
    + Returning > 0 from progress_callback will cancel the operation and will not update the ft_header will not be updated if cancelled.
    + recount_rows will return 0 on success or > 0 if cancelled or error.
    + Initial implementation is very simple and uses the existing PerconaFT cursor navigation. This may be inefficient due to needing to take individual dives (and locks) down the tree for each key.
    + Resulting row count will be (more) inaccurate if tree has unapplied messages for any of the following conditions (which can be rectified by running an OPTIMIZE TABLE before running a row recount):
      = INSERT messages that will be converted to UPDATE on apply due to key already existing. (upserts)
      = DELETE messages that get discarded in apply due to missing keys. (blind deletes)
      = Rows that need rollback applied for INSERT or DELETE. (unapplied rollback messages)
* Since the new counter extends the ft_header and bumps the version, there are data format upgrade/downgrade issues that need to be considered.
* A fair bit of code format fixes on any adjacent lines, mostly function signature reformatting to style guidelines and line length restriction to 80.

### DIFF
--- a/buildheader/make_tdb.cc
+++ b/buildheader/make_tdb.cc
@@ -510,8 +510,9 @@ static void print_db_struct (void) {
 			 "int (*update_broadcast)(DB *, DB_TXN*, const DBT *extra, uint32_t flags)",
 			 "int (*get_fractal_tree_info64)(DB*,uint64_t*,uint64_t*,uint64_t*,uint64_t*)",
 			 "int (*iterate_fractal_tree_block_map)(DB*,int(*)(uint64_t,int64_t,int64_t,int64_t,int64_t,void*),void*)",
-                         "const char *(*get_dname)(DB *db)",
-                         "int (*get_last_key)(DB *db, YDB_CALLBACK_FUNCTION func, void* extra)",
+			 "const char *(*get_dname)(DB *db)",
+			 "int (*get_last_key)(DB *db, YDB_CALLBACK_FUNCTION func, void* extra)",
+			 "int (*recount_rows)(DB* db, int (*progress_callback)(uint64_t count, uint64_t deleted, void* progress_extra), void* progress_extra)",
 			 NULL};
     sort_and_dump_fields("db", true, extra);
 }
--- a/ft/CMakeLists.txt
+++ b/ft/CMakeLists.txt
@@ -36,6 +36,7 @@ set(FT_SOURCES
   ft-flusher
   ft-hot-flusher
   ft-ops
+  ft-recount-rows
   ft-status
   ft-test-helpers
   ft-verify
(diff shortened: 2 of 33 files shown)


georgelorchpercona:
jenkins run http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/39/

another run pending after this branch was rebased to upstream/master

http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/40/

georgelorchpercona:
Seems something odd happened in censos5 debug, a rebuild and it 'disappeared'...hmmm...

http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/41/

georgelorchpercona:
So I made some minor changes and pushed/overwrote the branch with them. New jenkins here http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/48/

PR #335, opened 2015-10-23
[DB-884] TokuDB crashes under sysbench in READ-COMMITTED mode ...

Jun-Yuan:
This pull request is probably not ready yet over but I would like to raise the issues and hear advice sooner than later.

(1) the bug fix and ctest passed locally without any problem.

(2) jenkin build(http://jenkins.percona.com/job/PerconaFT-param/45/) have irrelavant failures. These two tests are quite irrelavent to my changes. I suspect some timing condition on which these tests count on have been changed but it is sort of hard for me to narrow down the search for the root cause(especially when the errno does not match with the retval, why?), because I could not reproduce the bugs locally.

(3) this is an attempt to see if we should bind the hooks to deterministicall reproduce a race bug with "DEBUG" build flag, which means the test makes no sense for RELEASE build.

Feel free to comment. Thank you.


BohuTANG:
yes, we need a debug_sync signal as in tokudbbackup

PR #342, opened 2015-12-23
Baseline

kuszmaul:
Add an option to print two new columns (checkpoint and checkpoint_count) to the tsv and csv output of the threaded stress tests.

You can see insertions stalling on
      ./perf_insert.tdb --txn_size 10000 --num_put_threads 4  --envdir test --print_performance --perf_format tsv --num_seconds 1800 --print_checkpoint_column --no-print_thread_performance


georgelorchpercona:
Great to see you back Bradley.

I am out until 2016 so will not really be able to review until then.

First glance though seems that your git history isn't clean, PR should be based on a clean git clone of the master branch which means that the diff should only show your single contribution commit as diif from master. Upon merge git might reconcile the differences but seeing that github is showing a 5 commit difference from master it may not mean a clean history. I will take a closer look in Jan and follow up then.

PR #348, opened 2016-04-28
Some changes to build on MacOs X

vlad-lesin:
This is mostly build options changes and printf converting error fixes.

http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1120/
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/144/

### DIFF
--- a/cmake_modules/TokuSetupCompiler.cmake
+++ b/cmake_modules/TokuSetupCompiler.cmake
@@ -95,8 +95,10 @@ set_cflags_if_supported(
   -Wno-error=missing-format-attribute
   -Wno-error=address-of-array-temporary
   -Wno-error=tautological-constant-out-of-range-compare
+  -Wno-error=maybe-uninitialized
   -Wno-ignored-attributes
   -Wno-error=extern-c-compat
+  -Wno-pointer-bool-conversion
   -fno-rtti
   -fno-exceptions
   )
@@ -152,13 +154,18 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL Clang)
   set(CMAKE_C_FLAGS_RELEASE "-g -O3 ${CMAKE_C_FLAGS_RELEASE} -UNDEBUG")
   set(CMAKE_CXX_FLAGS_RELEASE "-g -O3 ${CMAKE_CXX_FLAGS_RELEASE} -UNDEBUG")
 else ()
+  if (APPLE)
+    set(FLTO_OPTS "-fwhole-program")
+  else ()
+    set(FLTO_OPTS "-fuse-linker-plugin")
+  endif()
   # we overwrite this because the default passes -DNDEBUG and we don't want that
-  set(CMAKE_C_FLAGS_RELWITHDEBINFO "-flto -fuse-linker-plugin ${CMAKE_C_FLAGS_RELWITHDEBINFO} -g -O3 -UNDEBUG")
-  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-flto -fuse-linker-plugin ${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -g -O3 -UNDEBUG")
-  set(CMAKE_C_FLAGS_RELEASE "-g -O3 -flto -fuse-linker-plugin ${CMAKE_C_FLAGS_RELEASE} -UNDEBUG")
-  set(CMAKE_CXX_FLAGS_RELEASE "-g -O3 -flto -fuse-linker-plugin ${CMAKE_CXX_FLAGS_RELEASE} -UNDEBUG")
-  set(CMAKE_EXE_LINKER_FLAGS "-g -fuse-linker-plugin ${CMAKE_EXE_LINKER_FLAGS}")
-  set(CMAKE_SHARED_LINKER_FLAGS "-g -fuse-linker-plugin ${CMAKE_SHARED_LINKER_FLAGS}")
+  set(CMAKE_C_FLAGS_RELWITHDEBINFO "-flto ${FLTO_OPTS} ${CMAKE_C_FLAGS_RELWITHDEBINFO} -g -O3 -UNDEBUG")
+  set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-flto ${FLTO_OPTS} ${CMAKE_CXX_FLAGS_RELWITHDEBINFO} -g -O3 -UNDEBUG")
+  set(CMAKE_C_FLAGS_RELEASE "-g -O3 -flto ${FLTO_OPTS} ${CMAKE_C_FLAGS_RELEASE} -UNDEBUG")
+  set(CMAKE_CXX_FLAGS_RELEASE "-g -O3 -flto ${FLTO_OPTS} ${CMAKE_CXX_FLAGS_RELEASE} -UNDEBUG")
+  set(CMAKE_EXE_LINKER_FLAGS "-g ${FLTO_OPTS} ${CMAKE_EXE_LINKER_FLAGS}")
+  set(CMAKE_SHARED_LINKER_FLAGS "-g ${FLTO_OPTS} ${CMAKE_SHARED_LINKER_FLAGS}")
 endif ()
 
 ## set warnings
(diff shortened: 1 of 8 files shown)


cmake_modules/TokuSetupCompiler.cmake
```diff
@@ -192,14 +199,18 @@ endif ()
 set(CMAKE_C_FLAGS "-Wall -Werror ${CMAKE_C_FLAGS}")
 set(CMAKE_CXX_FLAGS "-Wall -Werror ${CMAKE_CXX_FLAGS}")
 
+## It looks like there is no more need in these changes for at least
```
laurynas-biveinis: Suggest to remove not comment out (git history will help if it ever needs restoring)

cmake_modules/TokuSetupCompiler.cmake
```diff
@@ -192,14 +199,18 @@ endif ()
...
-    add_definitions(-stdlib=libc++)
-  endif ()
-endif ()
+#if (APPLE)
+#  if (CMAKE_GENERATOR STREQUAL Xcode)
+#    set(CMAKE_XCODE_ATTRIBUTE_CLANG_CXX_LIBRARY "libc++")
+#  else ()
+#    add_definitions(-stdlib=libc++)
+#  endif ()
+#endif ()
```
georgelorchpercona: Similar to Laurynas suggestion, I would just remove all of this commented code here.

vlad-lesin:
The commented code is removed.

ft/txn/txn_manager.cc
```diff
@@ -538,7 +538,7 @@ void toku_txn_manager_handle_snapshot_create_for_child_txn(
         inherit_snapshot_from_parent(txn);
     }
    
-    toku_debug_txn_sync(pthread_self());
+    toku_debug_txn_sync(reinterpret_cast<uint64_t>(pthread_self()));
```
laurynas-biveinis: IMHO this is not a correct fix: the issue is that FT code uses uint64_t type instead of portable pthread_t. According to the standard, pthread_t might be a struct and should be treated opaquely. If possible, the right fix would be to replace uin64_t thread-id holder types with pthread_t, to replace equality comparisons with pthread_equal calls etc.

vlad-lesin:
> IMHO this is not a correct fix: the issue is that FT code uses uint64_t type instead of portable pthread_t. According to the standard, pthread_t might be a struct and should be treated opaquely. If possible, the right fix would be to replace uin64_t thread-id holder types with pthread_t,  to replace equality comparisons with pthread_equal calls etc.

I absolutely agree with you. But such a work requires additional time (BTW I have already replaced one comparison with pthread_equal). We do not support Mac platform officially. At the other hand the only place where this "debug sync" is used is in the test. So I think the current fix is a good tradeoff between the time spent on it and the ability to build the project on MacOS.

laurynas-biveinis:
Vlad, it's almost there already, only a few prototypes need adjusting. I did not change reinterpret_cast for printing the thread id in the test program though, but that's a very isolated use and the types are correct throughout the library.

diff --git a/ft/txn/txn_manager.cc b/ft/txn/txn_manager.cc
index c370398..88eca36 100644
--- a/ft/txn/txn_manager.cc
+++ b/ft/txn/txn_manager.cc
@@ -47,10 +47,10 @@ Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
 #include "util/omt.h"
 //this is only for testing
 
-static void  (* test_txn_sync_callback) (uint64_t, void *) = NULL;
+static void  (* test_txn_sync_callback) (pthread_t, void *) = NULL;
 static void * test_txn_sync_callback_extra = NULL;
 
-void set_test_txn_sync_callback(void (*cb) (uint64_t, void *), void *extra) {
+void set_test_txn_sync_callback(void (*cb) (pthread_t, void *), void *extra) {
         test_txn_sync_callback = cb;
         test_txn_sync_callback_extra = extra;
 }
@@ -538,7 +538,7 @@ void toku_txn_manager_handle_snapshot_create_for_child_txn(
         inherit_snapshot_from_parent(txn);
     }
    
-    toku_debug_txn_sync(reinterpret_cast<uint64_t>(pthread_self()));
+    toku_debug_txn_sync(pthread_self());
     
     if (copies_snapshot) { 
 	if(!records_snapshot) 
diff --git a/ft/txn/txn_manager.h b/ft/txn/txn_manager.h
index 28fa1ac..7cdc52c 100644
--- a/ft/txn/txn_manager.h
+++ b/ft/txn/txn_manager.h
@@ -43,7 +43,7 @@ Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
 
 #include "ft/txn/txn.h"
 
-void set_test_txn_sync_callback(void (*) (uint64_t, void*), void*);
+void set_test_txn_sync_callback(void (*) (pthread_t, void*), void*);
 #define toku_test_txn_sync_callback(a) ((test_txn_sync_callback)? test_txn_sync_callback( a,test_txn_sync_callback_extra) : (void) 0)
 
 #if TOKU_DEBUG_TXN_SYNC
diff --git a/src/tests/test_db_rowcount.cc b/src/tests/test_db_rowcount.cc
index 066f8e3..c440bdc 100644
--- a/src/tests/test_db_rowcount.cc
+++ b/src/tests/test_db_rowcount.cc
@@ -427,11 +427,6 @@ static void test_insert_commit_delete_rollback(DB_ENV* env) {
 
     db->close(db, 0);
 }
-static inline uint64_t time_in_microsec() {
-    struct timeval t;
-    gettimeofday(&t, NULL);
-    return t.tv_sec * (1UL * 1000 * 1000) + t.tv_usec;
-}
 
 static int test_recount_insert_commit_progress(
     uint64_t count,
diff --git a/src/tests/txn_manager_handle_snapshot_atomicity.cc b/src/tests/txn_manager_handle_snapshot_atomicity.cc
index 666135d..30cc16d 100644
--- a/src/tests/txn_manager_handle_snapshot_atomicity.cc
+++ b/src/tests/txn_manager_handle_snapshot_atomicity.cc
@@ -92,11 +92,11 @@ struct start_txn_arg {
 
 static struct test_sync sync_s;
 
-static void test_callback(uint64_t self_tid, void * extra) {
+static void test_callback(pthread_t self_tid, void * extra) {
     pthread_t **p = (pthread_t **) extra;
     pthread_t tid_1 = *p[0];
     pthread_t tid_2 = *p[1];
-    assert(pthread_equal((pthread_t)self_tid, tid_2));
+    assert(pthread_equal(self_tid, tid_2));
     printf("%s: the thread[%" PRIu64 "] is going to wait...\n", __func__, reinterpret_cast<uint64_t>(tid_1));
     test_sync_next_state(&sync_s);
     sleep(3);
diff --git a/src/ydb.cc b/src/ydb.cc
index 61ce5a8..55da418 100644
--- a/src/ydb.cc
+++ b/src/ydb.cc
@@ -3148,7 +3148,7 @@ toku_test_get_latest_lsn(DB_ENV *env) {
     return rval.lsn;
 }
 
-void toku_set_test_txn_sync_callback(void (* cb) (uint64_t, void *), void * extra) {
+void toku_set_test_txn_sync_callback(void (* cb) (pthread_t, void *), void * extra) {
     set_test_txn_sync_callback(cb, extra);
 }
 
diff --git a/src/ydb.h b/src/ydb.h
index bd2902e..facbfdc 100644
--- a/src/ydb.h
+++ b/src/ydb.h
@@ -60,4 +60,4 @@ extern "C" uint64_t toku_test_get_latest_lsn(DB_ENV *env) __attribute__((__visib
 extern "C" int toku_test_get_checkpointing_user_data_status(void) __attribute__((__visibility__("default")));
 
 // test-only function
-extern "C" void toku_set_test_txn_sync_callback(void (* ) (uint64_t, void *), void * extra) __attribute__((__visibility__("default")));
+extern "C" void toku_set_test_txn_sync_callback(void (* ) (pthread_t, void *), void * extra) __attribute__((__visibility__("default")));

georgelorchpercona:
Good discussion Laurynas. I agree, if we can treat pthread_t properly without needing to re-write the library, then lets do it here and now while touching the code. The tests I am not so concerned with as long as they do execute correctly on linux x86_64 platforms while assuming that pthread_t is a uint64_t.

vlad-lesin:
> Vlad, it's almost there already, only a few prototypes need adjusting. I did not change reinterpret_cast for printing the thread id in the test program though, but that's a very isolated use and the types are correct throughout the library.

Thank you, Laurynas. I have included your fixes in the patch.

georgelorchpercona:
Looks good, merging...

PR #349, opened 2016-04-28
Replace error to warning if transparent huge pages are enabled

vlad-lesin:
1. Error handle callback now has one more parameter which signals about if the message is error or warning.

2. Call error handle hook with "warning" parameter if the error is TOKUDB_HUGE_PAGES_ENABLED.

http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1118/
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/141/


vlad-lesin:
The requirements are changed.

PR #351, opened 2016-06-02
FT-722 : PerconaFT ydb environment validation errors are horrible and…

georgelorchpercona:
… useless.

Changed logging in ydb environment validation function to print more useful context such as:

TokuDB: Fileops directory [tokudb.directory] is missing from [<dir name removed>]
TokuDB: Persistent environment is missing while looking for rollback cachefile [tokudb.rollback] in [<dir name removed>]
TokuDB: rollback cachefile [tokudb.rollback] is missing from [<dir name removed>]
TokuDB: Recovery log is missing (persistent environment information is present) while looking for recovery log files in [<dir name removed>]

### DIFF
--- a/src/ydb.cc
+++ b/src/ydb.cc
@@ -623,32 +623,39 @@ ydb_recover_log_exists(DB_ENV *env) {
 }
 
 // Validate that all required files are present, no side effects.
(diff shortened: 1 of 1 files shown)


georgelorchpercona:
Self merging as changes are trivial.

PR #352, opened 2016-06-09
dir_per_db directory layout is implemented.

vlad-lesin:
Setting dir_per_db option on enables a new feature in PerconaFT where new DB
names are parsed and if a leading "<dir_name>/" is found, the name will be split
at the "/" and the new file will be placed within the <data_dir>/<dir_name> and
named with the remainder of the name string.

For example: Creating a new DB with the name "foo/bar" will expect that the
directory <data_dir>/foo exists and will create the new file as
<data_dir>/foo/bar.<ext>

FT renaming now also renames files. The "rename" operation is transactional,
so recovery and rollback logging is implemented for it. "rename" operation
as well as "create" use naming convention corresponding to dir_per_db option
value. The convertion between too modes is possible with "rename" operation.

[dl.func-tests.txt](https://github.com/percona/PerconaFT/files/307022/dl.func-tests.txt)
[dl.stress-test.txt](https://github.com/percona/PerconaFT/files/307023/dl.stress-test.txt)


vlad-lesin:
The ability to create directories for ft-data-file paths was added. Functional and stress tests passed.

PR #353, opened 2016-06-14
- Implement custom red-black tree to track available holes in FT map …

georgelorchpercona:
…file.

- Replace array based liner hole search with red-black tree search for first fit.
- Implement new and update existing tests.


georgelorchpercona:
https://tokutek.atlassian.net/browse/DB-995
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1195/
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/213/

esmet:
Why write a custom red-black tree for this when there's already an OMT? Plus, I'd be pretty surprised if finding a free block shows up on perf(1)

Jun-Yuan:
It doesn't make much difference wheather it's a OMT or red black tree to
impl the algorithm as long as it's a balanced search tree, because the
internal operations of the tree have to be adapted for this algorithm. it's
not an easy template thing. I don't want to mess up OMT either.

And yes to finding an available block shows up as the hottest on perf(1),
though it's not the most common case, it did happen for small blocks.

georgelorchpercona:
Not just small block sizes, this shows as significant issue even with default node sizes on sufficiently large data set. Smaller block sizes as generally used on fast storage/SSD for better performance exacerbate the issue. We have a few well documented reports of perf issues analyzed down to block_allocator_strategy::first_fit. Internal analysis reveals all of the strategies suffer from the same issue, the linear search of an array with many locks held on a heavily fragmented file. Defragging is out of the question as it takes the dataset offline during the index rebuild and rapidly devolves back to the same state. This solution shows an incredible improvement on a contrived/scaled down test.

leifwalsh:
Jun, is this the design we talked about? Did you write that down somewhere?
I wish I remembered it, it was neat :)
On Wed, Jun 15, 2016 at 15:29 georgelorchpercona <notifications@github.com>
wrote:

> Not just small block sizes, this shows as significant issue even with
> default node sizes on sufficiently large data set. Smaller block sizes as
> generally used on fast storage/SSD for better performance exacerbate the
> issue. We have a few well documented reports of perf issues analyzed down
> to block_allocator_strategy::first_fit. Internal analysis reveals all of
> the strategies suffer from the same issue, the linear search of an array
> with many locks held on a heavily fragmented file. Defragging is out of the
> question as it takes the dataset offline during the index rebuild and
> rapidly devolves back to the same state. This solution shows an incredible
> improvement on a contrived/scaled down test.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/percona/PerconaFT/pull/353#issuecomment-226294337>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAC1RuKgT6ZSP3iM5epGjmaz0Qm16OJZks5qMFKPgaJpZM4I1u5->
> .
>
-- 
-- 
Cheers,
Leif

Jun-Yuan:
Hi Leif! Yes, it is.  In ft/seriaalize/rbtree_mhs.h I commented the basic algorithm briefly. I should have added the comments: "Leif is the master mind of the design; Jun is the coding monkey of implement". Unfortunately I have been too busy to polish the work. :(  (Many many thanks to George).

BohuTANG:
@Jun-Yuan 
laught on your comments

ft/serialize/block_allocator.h
```diff
@@ -43,172 +43,155 @@ Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
...
+        uint64_t _offset;
+        uint64_t _size;
+        BlockPair(uint64_t o, uint64_t s) : _offset(o), _size(s) {}
+        int operator<(const struct BlockPair &rhs) const {
+            return _offset < rhs._offset;
+        }
+        int operator<(const uint64_t &o) const { return _offset < o; }
+    };
+
+    enum EAllocationStrategy {
```
laurynas-biveinis: enum class?

ft/serialize/block_table.cc
```diff
@@ -359,73 +398,89 @@ bool block_table::_is_valid_blocknum(struct translation *t, BLOCKNUM b) {
...
-    //Free the old block if it is not still in use by the checkpoint in progress or the previous checkpoint
-    bool cannot_free = (bool)
-        ((!for_checkpoint && _translation_prevents_freeing(&_inprogress,   b, &old_pair)) ||
-         _translation_prevents_freeing(&_checkpointed, b, &old_pair));
-    if (!cannot_free && old_pair.u.diskoff!=diskoff_unused) {
-        _bt_block_allocator.free_block(old_pair.u.diskoff);
+    // Free the old block if it is not still in use by the checkpoint in
+    // progress or the previous checkpoint
+    bool cannot_free =
+        (bool)((!for_checkpoint &&
```
laurynas-biveinis: redundant cast

ft/serialize/block_table.cc
```diff
@@ -587,42 +680,46 @@ void block_table::allocate_blocknum(BLOCKNUM *res, FT ft) {
...
+    // If the size is 0, no disk block has ever been assigned to this blocknum.
     if (old_pair.size > 0) {
-        //Free the old block if it is not still in use by the checkpoint in progress or the previous checkpoint
-        bool cannot_free = (bool)
-            (_translation_prevents_freeing(&_inprogress,   b, &old_pair) ||
-             _translation_prevents_freeing(&_checkpointed, b, &old_pair));
+        // Free the old block if it is not still in use by the checkpoint in
+        // progress or the previous checkpoint
+        bool cannot_free =
+            (bool)(_translation_prevents_freeing(&_inprogress, b, &old_pair) ||
```
laurynas-biveinis: Likewise

ft/serialize/rbtree_mhs.cc
```diff
@@ -0,0 +1,840 @@
...
+
+#include "ft/serialize/rbtree_mhs.h"
+#include "portability/toku_assert.h"
+#include <iostream>
+#include <iomanip>
+
+namespace MhsRbTree {
+
+    Tree::Tree() : _root(NULL) {
+        _align = 1;  // we do not align
```
laurynas-biveinis: init in constructor initialiser list?

ft/serialize/rbtree_mhs.cc
```diff
@@ -0,0 +1,840 @@
...
+    }
+
+    Tree::Tree(uint64_t align) : _root(NULL) { _align = align; }
```
laurynas-biveinis: Likewise

ft/serialize/rbtree_mhs.h
```diff
@@ -0,0 +1,359 @@
...
+    // underflow. If you got a file offset larger than MHS_MAX_VAL, it is not
+    // a problem here. :-/  - JYM
+    class OUUInt64 {
+       public:
+        static const uint64_t MHS_MAX_VAL = 0xffffffffffffffff;
+        OUUInt64() : _value(0) {}
+        OUUInt64(uint64_t s) : _value(s) {}
+        bool operator<(const OUUInt64 &r) const {
+            if (_value == MHS_MAX_VAL && r.ToInt() == MHS_MAX_VAL)
+                assert(0);  // impossible case
```
leifwalsh: assert(0) isn't a great way to handle exceptional conditions. It makes it dangerous to use -DNDEBUG. Also you could assert that the negative of the if condition holds and that would be more readable. 

Recommend handling the error less violently or replacing assert(0) with abort() if you really really want to crash. Also, at least document *why* this case is impossible.

ft/serialize/block_allocator.h
georgelorchpercona: I'm not sure that is what you think it is, it is a diff oddity, the actual code looks like:
```
    struct BlockPair {
        uint64_t _offset;
        uint64_t _size;
        BlockPair(uint64_t o, uint64_t s) : _offset(o), _size(s) {}
        int operator<(const struct BlockPair &rhs) const {
            return _offset < rhs._offset;
        }
        int operator<(const uint64_t &o) const { return _offset < o; }
    };

    enum EAllocationStrategy {
        BA_STRATEGY_FIRST_FIT = 1,
        BA_STRATEGY_BEST_FIT,
        BA_STRATEGY_PADDED_FIT,
        BA_STRATEGY_HEAT_ZONE
    };
```

laurynas-biveinis: I only have in mind "s/enum/enum class" here for a C++11 type-safe enum

georgelorchpercona: Ack, true, good point.

PR #354, opened 2016-06-22
AArch64 portability fixes

akopytov:
A set of fixes to make PerconaFT build and run on the AArch64 (64-bit ARMv8) architecture.

### DIFF
--- a/portability/toku_time.h
+++ b/portability/toku_time.h
@@ -98,9 +98,17 @@ double tokutime_to_seconds(tokutime_t)  __attribute__((__visibility__("default")
 
 // Get the value of tokutime for right now.  We want this to be fast, so we expose the implementation as RDTSC.
 static inline tokutime_t toku_time_now(void) {
+#if defined(__x86_64__) || defined(__i386__)
     uint32_t lo, hi;
     __asm__ __volatile__ ("rdtsc" : "=a" (lo), "=d" (hi));
     return (uint64_t)hi << 32 | lo;
+#elif defined (__aarch64__)
+    uint64_t result;
+    __asm __volatile__ ("mrs %[rt], cntvct_el0" : [rt] "=r" (result));
+    return result;
+#else
+#error No timer implementation for this platform
+#endif
 }
 
 static inline uint64_t toku_current_time_microsec(void) {
--- a/ft/serialize/compress.cc
+++ b/ft/serialize/compress.cc
@@ -235,7 +235,7 @@ void toku_decompress (Bytef       *dest,   uLongf destLen,
         strm.zalloc = Z_NULL;
         strm.zfree = Z_NULL;
         strm.opaque = Z_NULL;
-        char windowBits = source[1];
+        int8_t windowBits = source[1];
         int r = inflateInit2(&strm, windowBits);
         lazy_assert(r == Z_OK);
         strm.next_out = dest;
--- a/portability/tests/test-max-data.cc
+++ b/portability/tests/test-max-data.cc
@@ -64,7 +64,7 @@ int main(int argc, char *const argv[]) {
     if (verbose) printf("maxdata=%" PRIu64 " 0x%" PRIx64 "\n", maxdata, maxdata);
 
     // check the data size
-#if __x86_64__
+#if defined(__x86_64__) || defined(__aarch64__)
     assert(maxdata > (1ULL << 32));
 #elif __i386__
     assert(maxdata < (1ULL << 32));
--- a/src/tests/recovery_fileops_unit.cc
+++ b/src/tests/recovery_fileops_unit.cc
@@ -156,7 +156,7 @@ do_args(int argc, char * const argv[]) {
         choices[i] = -1;
     }
 
-    char c;
+    int c;
     while ((c = getopt(argc, argv, "vqhcrO:A:B:C:D:E:F:G:H:I:X:")) != -1) {
 	switch(c) {
         case 'v':
--- a/third_party/xz-4.999.9beta/build-aux/config.guess
+++ b/third_party/xz-4.999.9beta/build-aux/config.guess
@@ -1,10 +1,10 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
 #   Copyright (C) 1992, 1993, 1994, 1995, 1996, 1997, 1998, 1999,
-#   2000, 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008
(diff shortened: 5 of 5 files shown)


prohaska7:
Did you consider using clock_gettime CLOCK_REALTIME instead of the various low level processor instructions?  The realtime clock on linux is very fast since it memory maps the clock info into the user process space (see vdso).

akopytov:
Hi Rich,

I didn't, I was just adding a platform-specific implementation for toku_time_now() which uses RDTSC on x86.

But I see a couple of issues with clock_gettime(CLOCK_REALTIME):

- VDSO is not always available (it isn't on my hardware)
- nothing really beats a single CPU instruction in terms of overhead, and comments for toku_time_now() suggest it is supposed to be a low-overhead timer.

I did not measure overhead for instructions reading CPU cycles information vs. clock_gettime(). However, PERFORMANCE_SCHEMA uses the 'cycles' timer in contexts where it needs the lowest overhead timer and falls back to the nanoseconds (usually clock_gettime()-based) timer when the cycles one is not available.

"Additional Comments" in mysys/my_rdtsc.c is also an interesting read.

prohaska7:
What hardware are you running on?  I thought that vdso was a software feature in linux that makes gettimeofday (and other clock related system calls) fast.  It does this my avoid kernel system calls.

akopytov:
Sorry, I spoke too soon. VDSO is available on my platform, but whether it is actually used to accelerate each particular syscall depends on glibc, and that itself depends on hardware capabilities. I've seen certain system calls like sched_getcpu() using VDSO on x86, but not on my ARM-based platform.

It's hard to figure out from the glibc code whether VDSO is actually used for clock_gettime() in my case.

I just decided to compare overhead of a RDTSC-like timer and clock_gettime(), but noticed that code in mysys/my_rdtsc.c already does that, and that information is exposed through PERFORMANCE_SCHEMA. This is what I got:
```
mysql> select * from performance_timers;
+-------------+-----------------+------------------+----------------+
| TIMER_NAME  | TIMER_FREQUENCY | TIMER_RESOLUTION | TIMER_OVERHEAD |
+-------------+-----------------+------------------+----------------+
| CYCLE       |        99935323 |                1 |              0 |
| NANOSECOND  |      1000000000 |                1 |              5 |
| MICROSECOND |         1000000 |                1 |              6 |
| MILLISECOND |            1037 |                1 |              6 |
| TICK        |             103 |                1 |             58 |
+-------------+-----------------+------------------+----------------+
5 rows in set (0.00 sec)
```

If I read the code in my_rdtsc.c correctly, the above means that clock_gettime() is about 5 'cycle' units more expensive than the CYCLE timer itself. A 'cycle' in my case has nothing to do with CPU cycles, but is about 10 nanoseconds based on frequency, which means clock_gettime() has about 50 nanoseconds higher overhead as compared to the RDTSC-like timer.

georgelorchpercona:
Hey Alexey,
I'd be glad to test and merge your changes, but I need you to alter your commit message to include our GPL-2 license release as mentioned here https://github.com/percona/PerconaFT#contributing

Thanks!

akopytov:
Hi George,

That link says I need to contribute under the BSD 2-Clause license, not GPL-2. Also, do I really need to include the license text into each of the 5 commits, or just changing the pull request comment for all commits is enough?

georgelorchpercona:
Good question, the lawyers had us change the contribution requirement to the BSD 2-Clause as defined in the template here https://opensource.org/licenses/BSD-2-Clause/

I haven't reviewed your contribution yet and didn't notice the 5 different commits and discreet fixes. We don't get a lot of contributions so are figuring this out as they come. Usually I would ask that the commits get squashed, but these all look like unrelated changes towards your goal of AArch64 compilation, which makes it reasonable to keep them apart.

In this case, I believe you have two choices, one is to go back and change all of the commit messages, the other and easier choice is to put the license text into the initial pull request description/comment that covers all 5 commits.

akopytov:
Thanks for clarifications. I updated all 5 commits with the BSD 2-Clause license text.

portability/toku_time.h
```diff
@@ -98,9 +98,17 @@ double tokutime_to_seconds(tokutime_t)  __attribute__((__visibility__("default")
 
 // Get the value of tokutime for right now.  We want this to be fast, so we expose the implementation as RDTSC.
 static inline tokutime_t toku_time_now(void) {
+#if defined(x86_64) || defined(__i386__)
```
georgelorchpercona: Fails to build on all supported intel platforms (all 64 bit), this should probably be __x86_64__

```22:36:07 /mnt/workspace/PerconaFT-param/BUILD_TYPE/release/Host/ubuntu-precise-64bit/portability/toku_time.h:110:2: error: #error No timer implementation for this platform
22:36:07  #error No timer implementation for this platform
22:36:07   ^
22:36:07 /mnt/workspace/PerconaFT-param/BUILD_TYPE/release/Host/ubuntu-precise-64bit/portability/toku_time.h: In function ‘tokutime_t toku_time_now()’:
22:36:07 /mnt/workspace/PerconaFT-param/BUILD_TYPE/release/Host/ubuntu-precise-64bit/portability/toku_time.h:112:1: error: no return statement in function returning non-void [-Werror=return-type]
22:36:07  }
22:36:07  ^
22:36:07 cc1plus: all warnings being treated as errors```

see http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/54/

akopytov:
Fixed x86_64 builds by replacing `x86_64` with `__x86_64__` in `#ifdef`.

portability/toku_time.h
akopytov: Fixed. Thanks.

georgelorchpercona:
Thanks Alexey, this is testing much better.

georgelorchpercona:
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/55/

akopytov:
I'm not sure I read the Jenkins status above correctly. Is it good to merge, or some action required on my end?

georgelorchpercona:
Sorry Alexey, yes, that is the ctest suite which has recently become unstable due to some changes in jenkins. I also have pending PS jobs that will get logged here today. So far everything looks fine and no more is required from you.

georgelorchpercona:
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1211/
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/225/

akopytov:
Thanks, George!

PR #358, opened 2016-08-04
FT-732 : logical row counts inaccurate

georgelorchpercona:
- Added mechanism to track adjustments made to logical row counts per node and undo those adjustments if the node is evicted without being persisted.
- Fixed counting of updates that get turned into insert messages.
- Fixed stat64 test that was incorrectly changed during initial logical row count testing.
- Formatting changes for each function touched.


georgelorchpercona:
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/63/

PR #359, opened 2016-08-31
FT-735 : test-rbtree-insert-remove-without-mhs fails mem test and val…

georgelorchpercona:
…grind test

- Fixed test to properly initialize test data and free allocated resources.


georgelorchpercona:
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/64/

PR #360, opened 2016-09-14
fix bogus stalls in the lock tree for low concurrency applications

prohaska7:
1. Fix the release before wait race

The release before wait race occurs when a lock is released by transaction A after transaction B tried to acquire it but before transaction B has a chance to register it's pending lock request.  There are several ways to fix this problem, but we want to optimize for the common situation of minimal lock conflicts, which is what the lock acquisition algorithm currently does.  Our solution to the release before wait race is for transaction B to retry its lock request after its lock request has been added to the pending lock set.
1. Fix the retry race

The retry race occurs in the current lock retry algorithm which assumes that if some transaction is running lock retry, then my transaction does not also need to run it.  There is a chance that some pending lock requests will be skipped, but these lock requests will eventually time out.  For applications with small numbers of concurrent transactions, timeouts will frequently occur, and the application throughput will be very small.

The solution to the retry race is to use a group retry algorithm.  All threads run through the retry logic.  Sequence numbers are used to group retries into batches such that one transaction can run the retry logic on behalf of several transactions.  This amortizes the retry cost.  The sequence numbers also ensure that when a transaction releases its locks, all of the pending lock requests that it is blocking are retried.
1. Implement a mechanism to find and kill a pending lock request

Tags lock requests with a client id, use the client id as a key into the pending lock requests sets to find a lock request, complete the lock request with a lock timeout error.

Copyright (c) 2016, Rich Prohaska
All rights reserved.

2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.

### DIFF
--- a/buildheader/make_tdb.cc
+++ b/buildheader/make_tdb.cc
@@ -422,6 +422,7 @@ static void print_db_env_struct (void) {
                              "int (*set_checkpoint_pool_threads)(DB_ENV *, uint32_t)",
                              "void (*set_check_thp)(DB_ENV *, bool new_val)",
                              "bool (*get_check_thp)(DB_ENV *)",
+                             "void (*kill_waiter)(DB_ENV *, void *extra)",
                              NULL};
 
         sort_and_dump_fields("db_env", true, extra);
@@ -542,8 +543,8 @@ static void print_db_txn_struct (void) {
 	"int (*abort_with_progress)(DB_TXN*, TXN_PROGRESS_POLL_FUNCTION, void*)",
 	"int (*xa_prepare) (DB_TXN*, TOKU_XA_XID *, uint32_t flags)",
         "uint64_t (*id64) (DB_TXN*)",
-        "void (*set_client_id)(DB_TXN *, uint64_t client_id)",
-        "uint64_t (*get_client_id)(DB_TXN *)",
+        "void (*set_client_id)(DB_TXN *, uint64_t client_id, void *client_extra)",
+        "void (*get_client_id)(DB_TXN *, uint64_t *client_id, void **client_extra)",
         "bool (*is_prepared)(DB_TXN *)",
         "DB_TXN *(*get_child)(DB_TXN *)",
         "uint64_t (*get_start_time)(DB_TXN *)",
--- a/ft/txn/txn.cc
+++ b/ft/txn/txn.cc
@@ -269,6 +269,7 @@ static txn_child_manager tcm;
         .state = TOKUTXN_LIVE,
         .num_pin = 0,
         .client_id = 0,
+        .client_extra = nullptr,
         .start_time = time(NULL),
     };
 
@@ -705,12 +706,14 @@ bool toku_txn_has_spilled_rollback(TOKUTXN txn) {
     return txn_has_spilled_rollback_logs(txn);
 }
 
-uint64_t toku_txn_get_client_id(TOKUTXN txn) {
-    return txn->client_id;
+void toku_txn_get_client_id(TOKUTXN txn, uint64_t *client_id, void **client_extra) {
+    *client_id = txn->client_id;
+    *client_extra = txn->client_extra;
 }
 
-void toku_txn_set_client_id(TOKUTXN txn, uint64_t client_id) {
+void toku_txn_set_client_id(TOKUTXN txn, uint64_t client_id, void *client_extra) {
     txn->client_id = client_id;
+    txn->client_extra = client_extra;
 }
 
 time_t toku_txn_get_start_time(struct tokutxn *txn) {
--- a/ft/txn/txn.h
+++ b/ft/txn/txn.h
@@ -193,6 +193,7 @@ struct tokutxn {
     uint32_t num_pin; // number of threads (all hot indexes) that want this
                       // txn to not transition to commit or abort
     uint64_t client_id;
+    void *client_extra;
     time_t start_time;
 };
 typedef struct tokutxn *TOKUTXN;
@@ -293,8 +294,8 @@ void toku_txn_unpin_live_txn(struct tokutxn *txn);
 
 bool toku_txn_has_spilled_rollback(struct tokutxn *txn);
 
-uint64_t toku_txn_get_client_id(struct tokutxn *txn);
-void toku_txn_set_client_id(struct tokutxn *txn, uint64_t client_id);
+void toku_txn_get_client_id(struct tokutxn *txn, uint64_t *client_id, void **client_extra);
+void toku_txn_set_client_id(struct tokutxn *txn, uint64_t client_id, void *client_extra);
 
 time_t toku_txn_get_start_time(struct tokutxn *txn);
 
--- a/locktree/lock_request.cc
+++ b/locktree/lock_request.cc
@@ -65,6 +65,7 @@ void lock_request::create(void) {
     toku_cond_init(&m_wait_cond, nullptr);
 
     m_start_test_callback = nullptr;
+    m_start_before_pending_test_callback = nullptr;
     m_retry_test_callback = nullptr;
 }
 
@@ -79,7 +80,7 @@ void lock_request::destroy(void) {
 }
 
 // set the lock request parameters. this API allows a lock request to be reused.
-void lock_request::set(locktree *lt, TXNID txnid, const DBT *left_key, const DBT *right_key, lock_request::type lock_type, bool big_txn) {
+void lock_request::set(locktree *lt, TXNID txnid, const DBT *left_key, const DBT *right_key, lock_request::type lock_type, bool big_txn, void *extra) {
     invariant(m_state != state::PENDING);
     m_lt = lt;
     m_txnid = txnid;
@@ -91,6 +92,7 @@ void lock_request::set(locktree *lt, TXNID txnid, const DBT *left_key, const DBT
     m_state = state::INITIALIZED;
     m_info = lt ? lt->get_lock_request_info() : nullptr;
     m_big_txn = big_txn;
+    m_extra = extra;
 }
 
 // get rid of any stored left and right key copies and
@@ -173,6 +175,7 @@ int lock_request::start(void) {
         m_state = state::PENDING;
         m_start_time = toku_current_time_microsec() / 1000;
         m_conflicting_txnid = conflicts.get(0);
+        if (m_start_before_pending_test_callback) m_start_before_pending_test_callback();
         toku_mutex_lock(&m_info->mutex);
         insert_into_lock_requests();
         if (deadlock_exists(conflicts)) {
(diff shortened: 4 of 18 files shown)


vlad-lesin:
I could not contrive the case to repeat retry race, at this point the code looks good to me. But I have some concerns about possible contention on info->mutex in lock_request::retry_all_lock_requests(). I think we could create load test with some small range requests and big amount of queries, and then measure the average and peaks of query execution time to understand the effectiveness of the changes.

prohaska7:
Hello Vlad,
The following new tests from the pull request fail when run on the head of perconaft:

lock_request_start_release_wait asserts which demonstrates the race.

lock_request_start_retry_wait_race_2 sometimes hangs because of the race.

these tests pass with the changes from the pull request.

prohaska7:
the new test lock_request_start_retry_race_3 sometimes hangs on the head of the perconaft master because of the race on the should_retry_lock_requests variable and code in the retry function.  the pull request uses a group retry algorithm to solve this problem.

vlad-lesin:
Rich, thank you for your comments. I have reviewed the tests you mentioned. And I don't have doubts in correctness of both solutions: 1) retry lock request after lock request has been added to the pending lock set, 2) group retry algorithm.

But the general consern for me is info->mutex in lock_request::retry_all_lock_requests().

Let's consider the two variants of the code. The old code:
```c++
void lock_request::retry_all_lock_requests(locktree *lt) {
 ...
    if (!info->should_retry_lock_requests) {
        return;
    }

    toku_mutex_lock(&info->mutex);
...
    info->should_retry_lock_requests = false;
...
    toku_mutex_unlock(&info->mutex);
}
```

In this case if some thread retries all lock requests from pending set the other threads will not be blocked on  info->mutex because info->should_retry_lock_requests will be false.

The new code is the following:
```c++
void lock_request::retry_all_lock_requests(locktree *lt, void (*after_retry_all_test_callback)(void)) {
...
    info->retry_want++;
...
    if (info->pending_is_empty)
        return;

    toku_mutex_lock(&info->mutex);
...
    unsigned long long retry_gen = info->retry_want.load();
    if (retry_gen > info->retry_done) {
...
        info->retry_done = retry_gen;
    }

    toku_mutex_unlock(&info->mutex);
}
```

info->pending_is_empty becomes equal to true only when pending set is empty what means that while one thread retries lock requests from pending set the other threads will be blocked on info->mutex.

Let's look at this from another perspective. Assume we have N threads which have already incremented info->retry_want, but have not yet acquired lock on info->mutex. The pending set is not empty. One of these N threads takes lock on info->mutex and starts retry lock process, the other N-1 threads will be blocked on info->mutex. There can be situation when that thread have not yet finished retry locks process but another M threads try to release their lock ranges and hang on info->mutex too.

prohaska7:
Hello Vlad,
The races and the high level fixes are described in this doc http://prohaska7.blogspot.com/2016/09/races-in-fractal-tree-lock-manager.html.  The group retry algorithm ensures for all threads that the lock retries have been executed after the lock was released.  It does this in a way that groups together multiple threads.  Yes, multiple threads can be blocked on a mutex while some other threads is running the lock retries. These threads will check if the lock retries have been run on their behalf.

Here is an alternative group retry algorithm that uses a condition variable to synchronize many threads through the group retry logic.  I think it is clearer.  What do you think?

https://github.com/prohaska7/tokuft/commit/b30af2790957310642bfc43e85038d5dc5c35347

vlad-lesin:
Hello Rich.

Thank you for the fix, but I still have some concerns. I have illustrated the concerns here https://docs.google.com/presentation/d/1jmDNnW-0e3VmGtMblGEtFxxuMS56MF649a5-8UtqM3M . I considered 4 threads, one of which does locks retry, and the rest three are waiting on condition variable for the first one(see page 2), and then just exit(see page 3). Why they must wait if we already know that they will do nothing as retry_gen is bigger then their my_retry_lock(see page 1).

I propose the following fix:
```c++
--- a/locktree/lock_request.cc
+++ b/locktree/lock_request.cc
@@ -335,13 +335,12 @@ void lock_request::retry_all_lock_requests(locktree *lt, void (*after_retry_all_
     while (my_retry_want > info->retry_done) {
         if (!info->running_retry) {
             info->running_retry = true;
-            unsigned long long retry_gen = info->retry_want;
+           info->retry_done = info->retry_want;
             toku_mutex_unlock(&info->retry_mutex);
             retry_all_lock_requests_info(info);
             if (after_retry_all_test_callback) after_retry_all_test_callback();
             toku_mutex_lock(&info->retry_mutex);
             info->running_retry = false;
-            info->retry_done = retry_gen;
             toku_cond_broadcast(&info->retry_cv);
         } else {
             toku_cond_wait(&info->retry_cv, &info->retry_mutex);
```

I used sysbench to measure the performance of different variants. I generated table with 10k entries and used select_random_ranges.lua test. It generates "SELECT COUNT(*) WHERE k BETWEEN ? AND ? OR K BETWEEN ? AND ? ..." queries. The number of "k BETWEEN ? AND ?" is ten, each range is "100". The number of threads is 200. Here is the command line with parameters I use for sysbench:

```
src/sysbench ./src/lua/select_random_ranges.lua --mysql-db=test --mysql-user=root --mysql-password="" --mysql-storage-engine=TokuDB --delta=100 --threads=200 --time=60 run
```

Read-only load is used because of two reasons:

1) read requests are tranformed into write requests into the code:
```c++
// the locktree silently upgrades read locks to write locks for simplicity
int locktree::acquire_read_lock(TXNID txnid, const DBT *left_key, const DBT *right_key,
                               txnid_set *conflicts, bool big_txn) {
   return acquire_write_lock(txnid, left_key, right_key, conflicts, big_txn);
}
```

2) I would like to avoid io influence.

I tried to produce a contention on lock tree. And results are the following:

1) The variant without the changes:
```
SQL statistics:
    queries performed:
        read:                            3501060
        write:                           0
        other:                           0
        total:                           3501060
    transactions:                        3501060 (58314.75 per sec.)
    queries:                             3501060 (58314.75 per sec.)
    ignored errors:                      0      (0.00 per sec.)
    reconnects:                          0      (0.00 per sec.)

General statistics:
    total time:                          60.0373s
    total number of events:              3501060

Latency (ms):
         min:                                  0.13
         avg:                                  3.42
         max:                                178.81
         95th percentile:                      6.55
         sum:                            11990418.14

Threads fairness:
    events (avg/stddev):           17505.3000/3757.39
    execution time (avg/stddev):   59.9521/0.01
```

2) The variant for the initial fix (commit 05bcb0):
```
SQL statistics:
    queries performed:
        read:                            3464609
        write:                           0
        other:                           0
        total:                           3464609
    transactions:                        3464609 (57708.87 per sec.)
    queries:                             3464609 (57708.87 per sec.)
    ignored errors:                      0      (0.00 per sec.)
    reconnects:                          0      (0.00 per sec.)

General statistics:
    total time:                          60.0360s
    total number of events:              3464609

Latency (ms):
         min:                                  0.12
         avg:                                  3.46
         max:                                142.88
         95th percentile:                      6.91
         sum:                            11988633.26

Threads fairness:
    events (avg/stddev):           17323.0450/3842.69
    execution time (avg/stddev):   59.9432/0.01
```
As we can see, TPS is less then for variant (1).

3) Your last changes (commit 640339):
```
SQL statistics:
    queries performed:
        read:                            3481036
        write:                           0
        other:                           0
        total:                           3481036
    transactions:                        3481036 (57981.33 per sec.)
    queries:                             3481036 (57981.33 per sec.)
    ignored errors:                      0      (0.00 per sec.)
    reconnects:                          0      (0.00 per sec.)

General statistics:
    total time:                          60.0372s
    total number of events:              3481036

Latency (ms):
         min:                                  0.13
         avg:                                  3.44
         max:                                163.83
         95th percentile:                      6.79
         sum:                            11989654.33

Threads fairness:
    events (avg/stddev):           17405.1800/3876.65
    execution time (avg/stddev):   59.9483/0.01
```
TPS is better then for variant (2), but still not so good as variant (1).

4) Your last changes with my fix described above:
```
SQL statistics:
    queries performed:
        read:                            3482521
        write:                           0
        other:                           0
        total:                           3482521
    transactions:                        3482521 (58006.58 per sec.)
    queries:                             3482521 (58006.58 per sec.)
    ignored errors:                      0      (0.00 per sec.)
    reconnects:                          0      (0.00 per sec.)

General statistics:
    total time:                          60.0366s
    total number of events:              3482521

Latency (ms):
         min:                                  0.13
         avg:                                  3.44
         max:                                168.74
         95th percentile:                      6.79
         sum:                            11990587.58

Threads fairness:
    events (avg/stddev):           17412.6050/3831.79
    execution time (avg/stddev):   59.9529/0.01
```

The TPS is better then for (2) and (3), quite close to (1), but still not so good.

prohaska7:
Hello Vlad,
Your patch looks correct to me and it improves performance a bit, about 0.5% perf hit in transaction throughput compared to the current code.  The intent of this pull request is to fix a race condition WRT small numbers of running transactions, not improve overall transaction throughput.  New data structure would probably be required to improve performance.

vlad-lesin:
Thank you, Rich. Sorry for the long delays, now I have enough time for this task.

I have changed the code a little bit more, now the function looks the following:
```c++
void lock_request::retry_all_lock_requests(locktree *lt, void (*after_retry_all_test_callback)(void)) {
    lt_lock_request_info *info = lt->get_lock_request_info();

    // if there are no pending lock requests than there is nothing to do
    // the unlocked data race on pending_is_empty is OK since lock requests
    // are retried after added to the pending set.
    if (info->pending_is_empty)
        return;

    // get my retry generation (post increment of retry_want)
    unsigned long long my_retry_want = (info->retry_want += 1);

    toku_mutex_lock(&info->retry_mutex);

    // here is the group retry algorithm.
    // get the latest retry_want count and use it as the generation number of this retry operation.
    // if this retry generation is > the last retry generation, then do the lock retries.  otherwise,
    // no lock retries are needed.
    if ((my_retry_want - 1) == info->retry_done) {
        while (!info->pending_is_empty) {
            if (!info->running_retry) {
                info->running_retry = true;
                info->retry_done = info->retry_want;
                toku_mutex_unlock(&info->retry_mutex);
                retry_all_lock_requests_info(info);
                if (after_retry_all_test_callback) after_retry_all_test_callback();
                toku_mutex_lock(&info->retry_mutex);
                info->running_retry = false;
                toku_cond_broadcast(&info->retry_cv);
            } else {
                toku_cond_wait(&info->retry_cv, &info->retry_mutex);
            }
         }
    }
    toku_mutex_unlock(&info->retry_mutex);
}
```

There is no need for several threads to wait for the end of retry locks procedure, one thread is enough because even if several threads are waiting, only one will be working on locks retrying when the previous thread kick it.

How does it look for you?

The current code needs some minor changes as the library API has been changed. Would you mind if I create new branch, rebase your code on the newest master and make necessary changes to avoid build errors? And I would also add the fixes I proposed. There will be new pool request with the link to this pool request. Is it ok for you?

vlad-lesin:
For "kill query" there are the following pull requests:

https://github.com/percona/percona-server/pull/1738
https://github.com/percona/percona-server/pull/1739
https://github.com/percona/PerconaFT/pull/370

vlad-lesin:
For lock-tree-races fix see https://github.com/percona/PerconaFT/pull/371 .

PR #367, opened 2017-03-02
Added some new functions in API to edit directory file

vlad-lesin:
dirtool_attach - attach iname to dname, if dname exists - override the previous
                 value, add new record otherwise
dirtool_detach - remove record with corresponding dname, don't touch iname file
dirtool_move   - rename dname

See also:
https://github.com/percona/percona-server/pull/1458
https://github.com/percona/percona-server/pull/1459


georgelorchpercona:
I like this. Nice and clean. Also seems like a first step towards detachable/attachable tables, but that needs a lot more work to ensure recovery logs are applied and MVCC stacks on all rows are squashed before detach.

PR #368, opened 2017-03-14
Lp1671152

georgelorchpercona:
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/72/


georgelorchpercona:
Minor change, tested in production, self approving.

PR #372, opened 2017-05-19
TDB-2 : row counts are being lost intermittently after restarts

georgelorchpercona:
- Cleaned up access to BN logical_row_count_delta
- Moved lrc tracking back to individual basements and out of node itself due to
  need to track at the basement level for partial eviction.
- Fixed missing handling of row adjustment on partial eviction.
- Corrected logic in flush_callback to only adjust when exactly appropriate
  conditions.
- Corrected places to clear lrc adjustment counters when node gets written.

### DIFF
--- a/ft/ft-ops.cc
+++ b/ft/ft-ops.cc
@@ -651,8 +651,12 @@ void toku_ftnode_clone_callback(void *value_data,
     // set new pair attr if necessary
     if (node->height == 0) {
         *new_attr = make_ftnode_pair_attr(node);
-        node->logical_rows_delta = 0;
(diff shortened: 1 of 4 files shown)


georgelorchpercona:
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/79/
http://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/1904/
http://jenkins.percona.com/view/5.7/job/mysql-5.7-param/949/

georgelorchpercona:
Self approving

PR #375, opened 2017-06-15
TDB-23  TDB-24

georgelorchpercona:
TDB-23 : PerconaFT can assert when opening a dictionary with no useful information to error log
TDB-24 : PerconaFT asserts for various reasons deserializing nodes with no useful error output

### DIFF
--- a/ft/cachetable/cachetable.cc
+++ b/ft/cachetable/cachetable.cc
@@ -464,7 +464,10 @@ int toku_cachetable_openf (CACHEFILE *cfptr, CACHETABLE ct, const char *fname_in
 
 char *
 toku_cachefile_fname_in_env (CACHEFILE cf) {
-    return cf->fname_in_env;
+    if (cf) {
+        return cf->fname_in_env;
+    }
+    return nullptr;
 }
 
 void toku_cachefile_set_fname_in_env(CACHEFILE cf, char *new_fname_in_env) {
@@ -2888,6 +2891,10 @@ void *toku_cachefile_get_userdata(CACHEFILE cf) {
 CACHETABLE
 toku_cachefile_get_cachetable(CACHEFILE cf) {
     return cf->cachetable;
+}
+
+CACHEFILE toku_pair_get_cachefile(PAIR pair) {
+    return pair->cachefile;
 }
 
 //Only called by ft_end_checkpoint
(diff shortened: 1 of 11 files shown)


georgelorchpercona:
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/88/

ft/cachetable/cachetable.cc
```diff
@@ -464,7 +464,10 @@ int toku_cachetable_openf (CACHEFILE *cfptr, CACHETABLE ct, const char *fname_in
 
 char *
 toku_cachefile_fname_in_env (CACHEFILE cf) {
-    return cf->fname_in_env;
+    if (cf) {
+        return cf->fname_in_env;
+    }
+    return NULL;
```
laurynas-biveinis: s/NULL/nullptr/g ?

ft/ft.cc
```diff
@@ -435,7 +435,8 @@ int toku_read_ft_and_store_in_cachefile (FT_HANDLE ft_handle, CACHEFILE cf, LSN
     }
 
     int fd = toku_cachefile_get_fd(cf);
-    int r = toku_deserialize_ft_from(fd, max_acceptable_lsn, &ft);
+    const char *fn = toku_cachefile_fname_in_env(cf);
```
laurynas-biveinis: What's the policy on using "auto"?

tools/ftverify.cc
```diff
@@ -325,7 +325,7 @@ check_block(BLOCKNUM blocknum, int64_t UU(blocksize), int64_t UU(address), void
         }
         just_decompress_sub_block(&sb);
 	
-        r = verify_ftnode_sub_block(&sb);
+        r = verify_ftnode_sub_block(&sb, NULL, blocknum);
```
georgelorchpercona: nullptr

ft/cachetable/cachetable.cc
georgelorchpercona: Tough old habit to break ;)

ft/ft.cc
georgelorchpercona: "auto" only started creeping into FT just before we took ownership and there really isn't a defined policy other than Google code style guidelines which are somewhat ambivalent w.r.t. "auto". I'm not particularly thrilled with the idea of it in this use case as IMHO it decreases readability by hiding the nature of the true type and forces an observer to go look up the function signature if they need to understand what the type is and what they can do with it.

georgelorchpercona:
Made nullptr changes, re-jenkins at http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/93/
Merging...

PR #380, opened 2017-08-24
TDB-75: PerconaFT does not build standalone or with -DBUILD_TESTING=ON

vlad-lesin:
The problem is that some tests are not linked to liblocktree, and must not be
linked logically, because the structure of the code assumes that some directory
in the code contains the code for the separate library, and the tests inside
such directory test the corresponding library. At the other hand liblocktree
contained the definition of some PFS keys which were used inside of locktree
code, but that keys are initialized in ft-ops.cc which is the part of libft
library. So when libft is used without liblocktree there are "undefined symbol"
errors.

The solution is to move that keys definition from locktree.cc to ft-ops.cc.


georgelorchpercona:
Jenkins ctest http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/102/

PR #382, opened 2017-08-25
TDB-74 : PerconaFT encodes already encoded database name for director…

georgelorchpercona:
TDB-74 : PerconaFT encodes already encoded database name for directory name
  - PerconaFT encodes all single or runs of non alphanumeric characters in a
    dictionary name into a single underscore ('_'). When dir-per-db is in use,
    this extends to the target directory name for the 'database'.
  - This fix, prevents this transformation on the 'directory' portion of the
    dictionary name when dir-per-db is enabled and assumes that the caller has
    already scrubbed and encoded the name in a way that is safe for the underlying
    filesystem.

### DIFF
--- a/src/ydb_db.cc
+++ b/src/ydb_db.cc
@@ -83,53 +83,56 @@ ydb_db_layer_get_status(YDB_DB_LAYER_STATUS statp) {
     *statp = ydb_db_layer_status;
 }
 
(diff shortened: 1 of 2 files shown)


georgelorchpercona:
http://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/103/

vlad-lesin:
LGTM

PR #386, opened 2017-11-09
tests: compile error on arches where char is unsigned

grooverdan:
As the signed nature of char isn't specified in the C standard and we are using in a signed context here I'm making this explicit.

Compile error was:

src/tests/recovery_fileops_unit.cc:162:70: error: comparison is always true due to limited range of data type [-Werror=type-limits]
     while ((c = getopt(argc, argv, "vqhcrO:A:B:C:D:E:F:G:H:I:J:X:")) != -1) {
            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
cc1plus: all warnings being treated as errors

$CXX --version (Power 8 Architecture)
g++ (GCC) 6.3.1 20170515

### DIFF
--- a/src/tests/recovery_fileops_unit.cc
+++ b/src/tests/recovery_fileops_unit.cc
@@ -158,7 +158,7 @@ do_args(int argc, char * const argv[]) {
         choices[i] = -1;
     }
 
-    char c;
+    signed char c;
     while ((c = getopt(argc, argv, "vqhcrO:A:B:C:D:E:F:G:H:I:J:X:")) != -1) {
         switch (c) {
             case 'v':


grooverdan:
Thanks @georgelorchpercona.

prohaska7:
getopt returns an int.  the implicit case to char (original) or signed char (this change) is unnecessary for this use case, so why do it?

prohaska7:
why not just change type from 'signed char' to 'int'?  it would remove the question of why 'signed char' rather than 'int'.

grooverdan:
The returned value is compared against -1 and when char is treated as unsigned (allowed by the C standard), its value can never be -1.

While 'int' is obviously perfectly valid, and perhaps obvious looking at the examples in the getopt man page, I was a) keeping the change as minimal as possible, b) providing enlightenment to the nature of architecture differences (and odd/under-specified parts of the C standard) by example, and c) I probably just made it compile and work too.

Hopefully the commit message is obvious to anyone who questions why signed char and not char.

Why not 'int' now? Still a good question. Don't have a good answer. Apparently it was originally int and got changed to char in af3b8002786506fb3a042cefac14713d640f900d and I'm not really sure why it changed there too :-).

prohaska7:
The code at the beginning of time (afbea5d) uses 'char' type for the getopt return value.  This is incorrect.
Change 4bc27c8 changed it to 'int' which matches getopt return type and should be correct on both intel and power arch.
Change af3b800 changed it back to 'char '.  I suspect that this change was made by accident when a large development branch was merged.
Change 37101c0 changed it to 'signed char', which works on both intel and power.
I think that we should use the change from af3b800 as it also works on intel and power and is simpler (IMO).

PR #388, opened 2017-11-09
Add POWER implementation for toku_time_now

grooverdan:
One of the big gaps in Power enabled for tokudb was this function being missing.


grooverdan:
Thanks @georgelorchpercona. Much appreciated.

PR #389, opened 2017-11-09
test-max-data: ppc64 behaves like x86_64 and aarch64

grooverdan:
easy test case fix.


grooverdan:
thanks for the merge.

PR #390, opened 2017-11-09
Determine CPU level 1 cache line size at compile time

grooverdan:
Add CMake auto detection of the CPU level 1, data cache line size for OS-X and Linux and fall back to static hardware for known s390, Power64, Arm64 and default to 64 for x86/86_64 and others.

test_partitioned_counter isn't fixed yet for cache line size != 64


BohuTANG:
Hello @grooverdan ,

Since the machine of compling and running their CPU may not be the same, so this  patch is not suiteable for all the cases.

georgelorchpercona:
Compile time hardware check not runtime.

grooverdan:
> Compile time hardware check not runtime.

The cache line size of a CPU doesn't change during runtime, not on anything currently made anyway.

While a cross compile execution of sysctl/getconf will fail, the compile time of portability/toku_config.h.in will prevail as the macros used here are of the target of the cross compiler.

Happy to pull out the sysctl/getconf if that is better.

georgelorchpercona:
My point is that it is possible (but not necessarily probable) that the binaries can be cmake/compiled on a processor with a different cache line size than they will be run on. Our builds are performed on a cloud provider within virtualized environments. In theory, the cache line size should always be 64 but it is possible (but not probable) that the build time environment and the run time environment will be different.

The reality is that this PR ends up with no effect at all on TokuDB as the only code that is changed to utilize the new CPU_LEVEL1_DCACHE_LINESIZE outside of the tests is in the ftxx code which is not used by PerconaFT or TokuDB. ftxx is a C++ wrapper around the FT/bdb "C" API used by the Mongo* engines which are all deprecated.

grooverdan:
thanks for the explanation.

georgelorchpercona:
Not a problem @grooverdan Thank YOU for the PRs. I appreciate all contributions even if I can't get to reviewing, testing, and merging them right away.

PR #395, opened 2017-11-24
lp1733637, BLD-861: Fixed compilation issues with gcc 7 and Werror

dutow:
* Initialized some variables which were possibly uninitailized before
* Added fallthrough comments, or changed existing comments to the format gcc understands by default

### DIFF
--- a/ft/loader/loader.cc
+++ b/ft/loader/loader.cc
@@ -1078,7 +1078,7 @@ static void* extractor_thread (void *blv) {
     FTLOADER bl = (FTLOADER)blv;
     int r = 0;
     while (1) {
-        void *item;
+        void *item = nullptr;
         {
             int rq = toku_queue_deq(bl->primary_rowset_queue, &item, NULL, NULL);
             if (rq==EOF) break;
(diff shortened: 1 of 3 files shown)


ft/loader/loader.cc
```diff
@@ -1078,7 +1078,7 @@ static void* extractor_thread (void *blv) {
     FTLOADER bl = (FTLOADER)blv;
     int r = 0;
     while (1) {
-        void *item;
+        void *item = NULL;
```
laurynas-biveinis: nullptr ?

laurynas-biveinis: nullptr ?

dutow: Are we building mysql in C++11 mode? I tried to check that, and the only mention of that I found is that mysql8 will be built in c++11 mode. I don't see CMAKE_CXX_STANDARD mentioned anywhere in the repos and --std is only mentioned once with gnu99.

Now that I searched for it, nullptr is mentioned in our codebase, so it does work somehow.

laurynas-biveinis: PerconaFT is C++11 AFAIK

georgelorchpercona: Yes, TokuDB and PerconaFT require c++11

PR #396, opened 2017-12-19
thread sanitizer finds real bugs in fractal tree software

prohaska7:
The thread sanitizer finds thousands of issues (lock deadlocks, data races, etc) when running the fractal tree tests.  Luckily, there are only about a hundred unique issues since a couple of issues dominate the reporting.  While some of these issues were previously found with valgrind's data race detectors, the thread sanitizer identified new issues.  Many of these issues can be fixed by using atomic types in place of base types.  However, many of the issues have yet to be diagnosed.  These are discussed in the open issues section of this document.

The fractal tree software was developed on the Intel X86 machines, so there is a bias to its memory consistency model in the code. Some of the changes to the fractal tree softare to address issues found by the thread sanitizer are probably necessary to successfully run the fractal tree software on hardware architectures that have a memory consistency model that differs from the Intel X86 architecture.

Suppressions

The thread sanitizer allows one to suppress issues that either are not of interest right now or are known to be false positives.  These suppression rules are specified in a thread sanitizer suppression file.  Right now, there are several suppressions necessary to allow useful analysis of the remaining fractal tree software.  There are a few rules for unsafe operations that are thought to be acceptable. IMO, the use of these operations should be temporary until atomic variables are used or some other software change is made.

Fixed issues

* Many of the fractal tree tests had data races on some test specific global variables.  These races are easily fixed with atomic types in place of base types.

* Data races in the lock tree. Atomic types used to fix races on the current lock memory variable and the lock tree reference count variable.

* Data races in the cache table. Atomic types used to fix races on the current cache table size.

Open issues

* Lock order inversion in the concurrent lock tree.  When the concurrent lock tree gets rebalanced, the lock order from root to leaf changes.  This change is annotated for helgrind, so helgrind ignores any lock order inversions.  Since this issue occurs frequently, a suppression rule exists to ignore this issue.

* Data races with the in memory row count variable.  In particular, there are a couple of data races in the adjust logical row count function due to the fact that the in memory row count is updated with atomic functions but read normally.  Perhaps this function could use a atomic variable instead. Since this issue occurs frequenty, a suppression rule exists to ignore this issue until it can be investigated.

* Lock order inversion during recovery. Needs to be investigate. No suppression rule exists which makes analysis of the races difficult.

* Data race with fractal tree flow variables. Flow variable updates are atomic, reads are not.  Could be fixed with atomic variable type.

* Data race on rollback log detected between a client thread retiring a transaction and a hot indexer thread building a new index.  This looks like a serious bug.

* Data race on update of cachetable attributes.  Looks like a bug.

* Data race on fractal tree max MSN. Updates of the max MSN are atomic, reads are not. Could be fixed with atomic variable type.

* Data race on txn state.  Need to investigate impact of this race on correctness of code.

There will inevitably be more issue unearthed by the thread sanitizer as more tests are run.

Status variables

Most of the status variables are racy since we want lightweight overhead for maintaining them and are willing to live with some incorrect results.  The impact of using atomic types for all of the status variables should be investigated.

Tools

Ubuntu 17.10
Clang 4.0.1
Cmake 3.9.1

Directions

CXXFLAGS=-fsanitize=thread CC=clang CXX=clang++ cmake -D CMAKE_BUILD_TYPE=Debug ../tokuft
make -j8
TSAN_OPTIONS=”suppressions=tokuft.tsan.suppressions history_size=7 second_deadlock_stack=1” ctest -j8 -E ‘valgrind|helgrind|drd|try-’

Copyright (c) 2017, Rik Prohaska
All rights reserved.

### DIFF
--- a/ft/cachetable/cachetable-internal.h
+++ b/ft/cachetable/cachetable-internal.h
@@ -44,6 +44,7 @@ Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
 #include <util/kibbutz.h>
 #include <util/nb_mutex.h>
 #include <util/partitioned_counter.h>
+#include <atomic>
 
 //////////////////////////////////////////////////////////////////////////////
 //
@@ -498,12 +499,12 @@ private:
 
     pair_list* m_pl;
     cachefile_list* m_cf_list;
-    int64_t m_size_current;            // the sum of the sizes of the pairs in the cachetable
+    std::atomic_int64_t m_size_current;            // the sum of the sizes of the pairs in the cachetable
     int64_t m_size_cloned_data; // stores amount of cloned data we have, only used for engine status
     // changes to these two values are protected
     // by ev_thread_lock
     int64_t m_size_reserved;           // How much memory is reserved (e.g., by the loader)
-    int64_t m_size_evicting;           // the sum of the sizes of the pairs being written
+    std::atomic_int64_t m_size_evicting;           // the sum of the sizes of the pairs being written
 
     // these are constants
     int64_t m_low_size_watermark; // target max size of cachetable that eviction thread aims for
@@ -531,7 +532,7 @@ private:
     // in init, set to false during destroy
     bool m_run_thread;
     // bool that states if the eviction thread is currently running
-    bool m_ev_thread_is_running;
+    std::atomic_bool m_ev_thread_is_running;
     // period which the eviction thread sleeps
     uint32_t m_period_in_seconds;
     // condition variable on which client threads wait on when sleeping
--- a/cmake_modules/TokuSetupCompiler.cmake
+++ b/cmake_modules/TokuSetupCompiler.cmake
@@ -108,6 +108,8 @@ set_cflags_if_supported(
   -fno-rtti
   -fno-exceptions
   -Wno-error=nonnull-compare
+  -Wno-missing-braces
+  -Wno-address-of-packed-member
   )
 ## set_cflags_if_supported_named("-Weffc++" -Weffcpp)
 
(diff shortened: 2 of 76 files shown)


ft/cachetable/cachetable-internal.h
```diff
@@ -498,12 +499,12 @@ class evictor {
 
     pair_list* m_pl;
     cachefile_list* m_cf_list;
-    int64_t m_size_current;            // the sum of the sizes of the pairs in the cachetable
+    std::atomic_int64_t m_size_current;            // the sum of the sizes of the pairs in the cachetable
```
georgelorchpercona: This makes 'class evictor' not a POD and triggers static assertion via 'ENSURE_POD(evictor);' in cachetable.cc around line 3611

georgelorchpercona: Actually, all of these atomic_type_t specializations are incorrect as they do not exist in all supported back versions of glib, must use std::atomic<type> instead.

prohaska7: so for backward compatibility, use std::atomic<int> in place of std::atomic_int?

georgelorchpercona: yeah, these all need to be the older style specializations std::atomic<uint32_t>, etc... I easily changed those but the bigger issue is that by adding this to evictor (and possibly other class/structs) introduces non trivial layout and construction/destruction/copy semantics, which then causes failure of std::is_pod and triggers the static_assertion in ENSURE_POD. I do not recall offhand what the need for evictor and others to be POD was and what it would take to eliminate it.

Alternately, since the evictor is really just a singleton, a hack could be to just to move this to global scope, but that IMHO is going the wrong direction in the greater scheme to properly "C++-ize" the FT library.

PR #398, opened 2018-01-16
Fixing clang warnings/errors

dutow:
TDB-93 : Unaligned pointers in perconaFT
* Ignoring a warning about unaligned pointers
* Changed the zero cond initializer: C++ doesn't need a zero, that's C
* Disabled deprecation warnings for stdc functions with Clang: it doesn't allow different function declarations


georgelorchpercona:
https://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/110/

PR #409, opened 2018-03-29
[TDB-20] With checkpoint disabled, closing FT file still rotates header.

Jun-Yuan:
jenkin: https://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/120/
         [Updated the comments and formats from the last PR.]

         This is an attempt based on @BohuTang's patch
             https://github.com/percona/PerconaFT/pull/331

         in particular, the ctest mostly comes from his with minor updates.

         Summary:
             1) The issue is real.
             2) @Bohu's patch works for mediating the issue but somewhat introduces
                a ft leak that the ft with ref=0 may be left until next open/close or
                checkpoint.
             3) If ft_close should be held off upon a backup, then it simply means
                a backup should hold a ref to the ft and responsible for closing ft up
                if it is the last one to hold it, like any other possible referencer.
                including the checkpoint, the txn and ft handlers.
             4) mtr test contributed by @BohuTang still passes. (in a different PS PR).


BohuTANG:
Hello,

Any updates regarding this? 
Thanks.

PR #410, opened 2018-04-11
TDB-184 : Broken fsync error handling

georgelorchpercona:
- Factored out i/o error reporting to be common and try to obtain actual
  filename from descriptor.
- Fixed ENOSPC reporting to use common i/o reporting.
- Fixed fsync retry logic to retry EINTR, EIO, and ENOLCK based on InnoDB code
  and upstream mysql issues 34823 and 90296. Retry will print message every 100
  attempts in a loop.
- Touched up 'slow fsync' tracking and reporting to use new i/o error reporting.
- Fixed close retry logic to be same as fsync.


georgelorchpercona:
https://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/122/
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2147/
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1766/
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1771/ 

The second 5.7 jenkins is a matrix reload on 4 failures due to some odd git state weirdness, they failed to clone/checkout my branch.

portability/file.cc
```diff
@@ -72,67 +72,98 @@ void toku_fs_get_write_info(time_t *enospc_last_time, uint64_t *enospc_current,
...
+            buf);
+#else
+    fprintf(stderr,
+            "%.24s : PerconaFT I/O problem, fd[%d], file[?] : %s\n",
+            tstr,
+            fd,
+            buf);
+#endif
+    fflush(stderr);
+}
```
laurynas-biveinis: va_end

portability/file.cc
```diff
@@ -72,67 +72,98 @@ void toku_fs_get_write_info(time_t *enospc_last_time, uint64_t *enospc_current,
...
         }
-    }
-    default:
-	break;
+        case ENOSPC: {
+            if (toku_assert_on_write_enospc) {
+                report_io_problem(
+                    fd, "Failed write of [%" PRIu64 "] bytes.", (uint64_t)len);
+                int out_of_disk_space = 1;
+                assert(!out_of_disk_space);  // Give an error message that might
```
laurynas-biveinis: Huh. bool at least?

portability/file.cc
```diff
@@ -72,67 +72,98 @@ void toku_fs_get_write_info(time_t *enospc_last_time, uint64_t *enospc_current,
...
-                fprintf(stderr, "retry in %d second%s\n", toku_write_enospc_sleep, toku_write_enospc_sleep > 1 ? "s" : "");
-                fflush(stderr);
-            }
-            sleep(toku_write_enospc_sleep);
+        case EINTR: {  // The call was interrupted by a signal before any data
+                       // was written; see signal(7).
+            report_io_problem(fd,
+                              "Write of [%" PRIu64
+                              "] bytes interrupted.  Retrying.",
+                              (uint64_t)len);
```
laurynas-biveinis: C++ style cast

portability/file.cc
```diff
@@ -72,67 +72,98 @@ void toku_fs_get_write_info(time_t *enospc_last_time, uint64_t *enospc_current,
...
-            toku_sync_fetch_and_sub(&toku_write_enospc_current, 1);
             break;
         }
-    }
-    default:
-	break;
+        case ENOSPC: {
+            if (toku_assert_on_write_enospc) {
+                report_io_problem(
+                    fd, "Failed write of [%" PRIu64 "] bytes.", (uint64_t)len);
```
laurynas-biveinis: C++ style cast

portability/file.cc
```diff
@@ -72,67 +72,98 @@ void toku_fs_get_write_info(time_t *enospc_last_time, uint64_t *enospc_current,
...
+                int out_of_disk_space = 1;
+                assert(!out_of_disk_space);  // Give an error message that might
+                                             // be useful if this is the only one
+                                             // that survives.
+            } else {
+                toku_sync_fetch_and_add(&toku_write_enospc_total, 1);
```
laurynas-biveinis: Would it be too much work to convert to std::atomic?

portability/file.cc
```diff
@@ -597,20 +628,48 @@ int toku_os_close_with_source_location(
...
-                printf("rr=%d (%s)\n", rr, strerror(rr));
-            assert(rr == EINTR);
+        if (r == -1) {
+            int what_errno = get_error_errno();
+            switch (what_errno) {
+                case ENOLCK:
+                // fallthrough
+                case EINTR:
+                // fallthrough
+                case EIO:
```
laurynas-biveinis: But the issue being fixed here is that EIO must fail at once, exactly the opposite of EINTR/ENOLCK

portability/file.cc
```diff
@@ -708,17 +768,46 @@ void file_fsync_internal_with_source_location(int fd,
...
+            r = fsync(fd);
         }
-	if (r) {
-            assert(get_error_errno() == EINTR);
-            eintr_count++;
-	}
-    }
+
+        if (r == -1) {
+            int what_errno = get_error_errno();
```
laurynas-biveinis: Mostly duplicated code with the above

portability/file.cc
```diff
@@ -72,67 +72,98 @@ void toku_fs_get_write_info(time_t *enospc_last_time, uint64_t *enospc_current,
     *enospc_total = toku_write_enospc_total;
 }
 
+
+void report_io_problem(int fd, const char *fmt, ...) __attribute__((format(printf, 2, 3)));
```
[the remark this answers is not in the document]
georgelorchpercona: Ohh, good catch, I forgot attribute could be used like that.

portability/file.cc
georgelorchpercona: This was original FT code and I saw no reason to change it, funky as it is.

portability/file.cc
georgelorchpercona: similar, slightly modified FT code, but yeah

portability/file.cc
georgelorchpercona: probably not as these are pretty local to file.cc

portability/file.cc
laurynas-biveinis: oh, and static constexpr yada yada *if* you decide to touch it

portability/file.cc
georgelorchpercona: That is not what I came to understand about EIO, which could be incorrect. https://stackoverflow.com/questions/42434872/writing-programs-to-cope-with-i-o-errors-causing-lost-writes-on-linux

Otherwise, what is the point of this exercise at all where FT would generally assert on all errors that were not EINTR or ENOSPC.

laurynas-biveinis: That's what I was referring with my comment on the TDB issue: "I didn't realise that was a release-build assert too - in which case the bug is not nearly as serious"

georgelorchpercona:
OK, I reworked it a bit more and factored out the write error handing as well as it wasn't really doing what the function names might indicate. Here is a new set of jenkins runs. Please take another look and approach it as if it were a new review and not just the items commented.

https://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/123/
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2151/
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1772/

portability/file.cc
```diff
@@ -72,70 +80,124 @@ void toku_fs_get_write_info(time_t *enospc_last_time, uint64_t *enospc_current,
...
-                    fprintf(stderr, "%.24s PerconaFT No space when writing %" PRIu64 " bytes to %*s ", tstr, (uint64_t) len, (int) n, symname);
-		}
-                fprintf(stderr, "retry in %d second%s\n", toku_write_enospc_sleep, toku_write_enospc_sleep > 1 ? "s" : "");
-                fflush(stderr);
+static void __attribute__((format(printf, 2, 3)))
+report_io_problem(int fd, const char *fmt, ...) {
+    va_list ap;
+    va_start(ap, fmt);
+
+    constexpr int tstr_length = 26;
```
laurynas-biveinis: static constexpr (but only if there are other comments to address)

PR #413, opened 2018-05-09
the cachetable-simple-pin-nonblocking-cheap test occasionally fails due to a locking conflict with the cachetable evictor.

prohaska7:
The cachetable-simple-pin-noblocking-cheap test occasionally fails due to a TOKUDB_TRY_AGAIN error being kicked out of the cachetable to the test.  The test does not expect this error so it crashes.  The TOKUDB_TRY_AGAIN error was caused by a locking conflict between the test and the cachetable evictor which periodically runs on a background thread.  Since the evictor is not an important part of this test, its execution was removed by setting the cachetable size larger than the test's cachetable footprint.

Tools:
Ubuntu 18.04
gcc 7.3.0
Cmake 3.10.2

Copyright (c) 2018, Rik Prohaska
All rights reserved.

### DIFF
--- a/ft/tests/cachetable-simple-pin-nonblocking-cheap.cc
+++ b/ft/tests/cachetable-simple-pin-nonblocking-cheap.cc
@@ -107,7 +107,11 @@ run_case_that_should_fail(CACHEFILE f1, pair_lock_type first_lock, pair_lock_typ
 
 static void
 run_test (void) {
-    const int test_limit = 12;
+    // sometimes the cachetable evictor runs during the test.  this sometimes causes cachetable pair locking contention,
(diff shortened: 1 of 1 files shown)


georgelorchpercona:
Thanks again Rik!
https://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/126/

prohaska7:
This problem was a nuisance since it would occur on perhaps 10% of test runs.  First needed to find which cachetable code returned the TOKUDB_TRY_AGAIN. It was the maybe_pin_pair testing whether or not the pair lock was expensive.  Sometimes it was due to some other thread holding a write lock on the pair.  Who?  It was the evictor thread. So, turn off evictions by making the eviction trigger huge.

PR #415, opened 2018-08-10
PS-4732 : PerconaFT needs more 8.0 conditional compilation to pull in…

georgelorchpercona:
… proper pfs headers

- Added conditional >= 8.0 to bring in new headers in 8.0 that don't exist in
  lesser versions.

### DIFF
--- a/portability/toku_instr_mysql.h
+++ b/portability/toku_instr_mysql.h
@@ -12,8 +12,15 @@
 // undefine them here to avoid compilation errors.
 #undef __STDC_FORMAT_MACROS
 #undef __STDC_LIMIT_MACROS
-#include <mysql/psi/mysql_file.h>    // PSI_file
-#include <mysql/psi/mysql_thread.h>  // PSI_mutex
+#include "mysql/psi/mysql_file.h"    // PSI_file
+#include "mysql/psi/mysql_thread.h"  // PSI_mutex
+#include "mysql/psi/mysql_stage.h"   // PSI_stage
+
+#if (MYSQL_VERSION_ID >= 80000)
+#include "mysql/psi/mysql_cond.h"
+#include "mysql/psi/mysql_mutex.h"
+#include "mysql/psi/mysql_rwlock.h"
+#endif  //  (MYSQL_VERSION_ID >= nn)
 
 #ifndef HAVE_PSI_MUTEX_INTERFACE
 #error HAVE_PSI_MUTEX_INTERFACE required


georgelorchpercona:
https://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/129/
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2202/
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1898/

portability/toku_instr_mysql.h
```diff
@@ -12,8 +12,16 @@
...
 #undef __STDC_LIMIT_MACROS
-#include <mysql/psi/mysql_file.h>    // PSI_file
-#include <mysql/psi/mysql_thread.h>  // PSI_mutex
+#include "mysql/psi/mysql_file.h"    // PSI_file
+#include "mysql/psi/mysql_thread.h"  // PSI_mutex
+
+#if (MYSQL_VERSION_ID >= 80000)
+#include "mysql/psi/mysql_cond.h"
+#include "mysql/psi/mysql_memory.h"
+#include "mysql/psi/mysql_mutex.h"
```
laurynas-biveinis: mysql_memory.h and mysql_stage.h are present in 5.7, thus their include should not depend on version >= 8.0 ?

georgelorchpercona: Perhaps.  5.6 does not have mysql_memory.h so there would need to be an additional conditional for >= 50700 and both 5.6 and 5.7 do not explicitly need these as they are pulled in through some other header.

georgelorchpercona:
OK, made some minor changes to the conditional.  New jenkins just for safety :
https://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/130/
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2204/
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1901/

portability/toku_instr_mysql.h
```diff
@@ -12,8 +12,19 @@
...
+#include "mysql/psi/mysql_thread.h"  // PSI_mutex
+#include "mysql/psi/mysql_stage.h"   // PSI_stage
+
+#if (MYSQL_VERSION_ID >= 80000)
+#include "mysql/psi/mysql_cond.h"
+#include "mysql/psi/mysql_mutex.h"
+#include "mysql/psi/mysql_rwlock.h"
+
+#elif (MYSQL_VERSION_ID >= 50700)
+#include "mysql/psi/mysql_memory.h"
```
laurynas-biveinis: It looks like mysql_memory.h will not be included in 8.0 now?

laurynas-biveinis: It looks now mysql_memory.h will not be included in 8.0? The checks against 8.0 and 5.7 have to be independent

georgelorchpercona: Yeah, you're correct.  Weird, it passed all build tests, so perhaps mysql_memory.h isn't even needed here...

georgelorchpercona: OK, so mysql_memory.h isn't required at all.  This makes some sense as the FT PFS implementation was initially done against 5.6 which did not have memory instrumentation and we never went back to add it in 5.7.  Fixing...

georgelorchpercona: Pushed new changes but no new jenkins as jenkins seems unable to schedule new jobs at the moment.  It did pass compilation on 5.6, 5.7, and 8.0.12 locally :-\

PR #417, opened 2018-08-16
PS-4381 : PerconaFT needs different libraries to link to 8.0 debug sy…

georgelorchpercona:
…nc and PFS

- Fixed another new split out of libs with strange cross dependencies that for
  some unknown reason causes problems with linkage of supporting tools that are
  only used within the ct used within the cest suite.est suite.


georgelorchpercona:
https://jenkins.percona.com/view/TokuDB/job/PerconaFT-param/135/
https://jenkins.percona.com/view/PS%205.6/job/percona-server-5.6-param/2209/
https://jenkins.percona.com/view/5.7/job/mysql-5.7-param/1907/

georgelorchpercona:
Self approving as changes are similar and minor to previous change and inconsequential to the FT library itself, only support tools used in ctest

PR #420, opened 2018-10-25
Ps 4960

georgelorchpercona:
First commit is purely a whole file clang-format of omt.cc as it was completely different formatting from the standard.  Second commit is actual code fixes for PS-4960


georgelorchpercona:
Jenkins, if the job still works...jenkins against 5., 5.7, 8.0 will come later.
https://ps.cd.percona.com/view/0ther/job/PerconaFT-param/9/

PR #423, opened 2019-02-07
PS-5406 - jemalloc 5.0.x causes TokuDB/PerconaFT to assert on load

georgelorchpercona:
- jemalloc mallctl_f("opt.lg_chunk", ...) is failing in memory init with
  jemalloc 5.1.0.  This option was removed in 5.0.0 and versions of jemalloc
  greater than 5.0 no longer operate with the concept of lg_chunk.  There is no
  equivalent replacement functionality.  For this specific issue, lg_chunk is
  used by PerconaFT to set the threshold where allocations will switch to using
  mmap instead of the standard allocator.  It is reasonable to recognize the
  lg_chunk query failure and pick a rational value and continue on rather than
  failing and asserting.  For the older jemalloc versions, this value was 4MB
  by default.

- Fixed missing initialization of global flag indicating memory startup status.

- Added more assertions in the startup path to catch specific issues.

- Altered behavior of jemalloc startup to recognize when lg_chunk is not
  supported and use 4MB instead.

### DIFF
--- a/ft/ft-ops.cc
+++ b/ft/ft-ops.cc
@@ -4973,6 +4973,7 @@ int toku_ft_layer_init(void) {
 
     // Portability must be initialized first
     r = toku_portability_init();
+    assert(r==0);
     if (r) {
         goto exit;
     }
@@ -4980,6 +4981,7 @@ int toku_ft_layer_init(void) {
     toku_pfs_keys_init("fti");
 
     r = db_env_set_toku_product_name("tokudb");
+    assert(r==0);
     if (r) {
         goto exit;
     }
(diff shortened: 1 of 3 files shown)


georgelorchpercona:
https://ps.cd.percona.com/job/PerconaFT-param/1/
https://ps.cd.percona.com/view/5.6/job/percona-server-5.6-param/50/
https://ps.cd.percona.com/view/5.7/job/percona-server-5.7-param/112/
https://ps.cd.percona.com/view/8.0/job/percona-server-8.0-param/217/

Already discussed solution in JIRA issue and had preliminary review of branch, self approving.

PR #424, opened 2019-03-19
PS-5501 : Re-license PerconaFT 'locktree' to Apache V2

georgelorchpercona:
- Fixed some incomplete relicensed files from previous round.

- Added missing license text to some.

- Relicensed more files to Apache V2 that locktree depends on.


georgelorchpercona:
License text changes only, no jenkins required (compiled locally). Self approving.

PR #425, opened 2019-05-09
Correct width calculation in logger.cc

grooverdan:
From @chriscalender in https://github.com/MariaDB/server/pull/1169


svoj:
Proposed MariaDB change was incorrect. A proper fix and reasoning MariaDB/server@dd939d6f7e57e418a6f80fb9057eef8823beaff6.

PR #426, opened 2019-06-21
PS-5102 : Merge MySQL 8.0.16

georgelorchpercona:
- Fixed cmake error:
  CMake Error at storage/tokudb/PerconaFT/cmake_modules/TokuMergeLibs.cmake:10 (GET_TARGET_PROPERTY):
    get_target_property() called with non-existent target "-lpthread".
  Call Stack (most recent call first):
    storage/tokudb/PerconaFT/cmake_modules/TokuMergeLibs.cmake:38 (TOKU_GET_DEPENDEND_OS_LIBS)
    storage/tokudb/PerconaFT/portability/CMakeLists.txt:25 (toku_merge_static_libs)

- Fixed cmake error:
  CMake Error at storage/tokudb/PerconaFT/cmake_modules/TokuMergeLibs.cmake:10 (GET_TARGET_PROPERTY):
    get_target_property() called with non-existent target "-lpthread".
  Call Stack (most recent call first):
    storage/tokudb/PerconaFT/cmake_modules/TokuMergeLibs.cmake:38 (TOKU_GET_DEPENDEND_OS_LIBS)
    storage/tokudb/PerconaFT/src/CMakeLists.txt:29 (toku_merge_static_libs)

PR #427, opened 2019-07-01
PS-5102: Fix configuration error with MySQL 8.0.16

dutow:
"-lpthread" is passed as a dependency to to library merge functions,
causing a CMake error because of a non existent target. This patch
excludes any external library starting with "-l" from these checks.

### DIFF
--- a/cmake_modules/TokuMergeLibs.cmake
+++ b/cmake_modules/TokuMergeLibs.cmake
@@ -6,7 +6,7 @@ FUNCTION(TOKU_GET_DEPENDEND_OS_LIBS target result)
   IF(deps)
    FOREACH(lib ${deps})
     # Filter out keywords for used for debug vs optimized builds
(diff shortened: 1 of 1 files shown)


laurynas-biveinis:
@dutow, what is the relation, if any, between this PR and https://github.com/percona/PerconaFT/pull/426?

dutow:
Ops, I didn't realize that George already submitted a pull request about this part. It's for the same issue, so this isn't needed then.

laurynas-biveinis:
@dutow and @georgelorchpercona, you took different approaches, which one you think we should go with?

georgelorchpercona:
I think @dutow approach is probably more correct than mine.

dutow:
https://ps57.cd.percona.com/job/percona-server-5.7-param/81/
https://ps56.cd.percona.com/job/percona-server-5.6-param/15/

dutow:
Changed to an even simpler diff.

https://ps56.cd.percona.com/job/percona-server-5.6-param/16/
https://ps57.cd.percona.com/job/percona-server-5.7-param/82/

PR #428, opened 2019-07-03
PS-5102: Remove old build_jemalloc target

dutow:
Current cmake and build scripts refer to a "build_jemalloc" target,
which was never actually present in the git tree. Starting with
8.0.16, MySQL CMake scripts set CMP00046 to NEW, which emits errors
on non existing dependencies: the scripts work as before, and omit
a working makefile, but cmake reports an exit code of 1 in the end,
causing jenkins failures.

This change removes every mention of the build_jemalloc target,
fixing the issue.

### DIFF
--- a/cmake_modules/TokuThirdParty.cmake
+++ b/cmake_modules/TokuThirdParty.cmake
@@ -1,39 +1,5 @@
 include(ExternalProject)
 
-if (NOT DEFINED LIBJEMALLOC)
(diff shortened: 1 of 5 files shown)


dutow:
https://ps56.cd.percona.com/job/percona-server-5.6-param/20/
https://ps57.cd.percona.com/job/percona-server-5.7-param/91/

PR #430, opened 2019-09-05
PS-5932: Implementing --tokudb-force-recovery=6 to skip reading the logs

dutow:
This could be useful when the server crashed with a corrupted rollback file,
and is unable to start up. Specifying --tokudb--force-recovery=6 --super-read-only
should start it up in a read-only, but usable state. Some data may be lost and
unrecoverable.

Starting the server without the read only option is *NOT* supported.

### DIFF
--- a/ft/ft.h
+++ b/ft/ft.h
@@ -181,11 +181,11 @@ void tokuft_update_product_name_strings(void);
 extern char toku_product_name[TOKU_MAX_PRODUCT_NAME_LENGTH];
 
 struct toku_product_name_strings_struct {
-    char db_version[sizeof(toku_product_name) + sizeof("1.2.3 build ") + 256];
-    char environmentdictionary[sizeof(toku_product_name) + sizeof(".environment")];
-    char fileopsdirectory[sizeof(toku_product_name) + sizeof(".directory")];
-    char single_process_lock[sizeof(toku_product_name) + sizeof("___lock_dont_delete_me")];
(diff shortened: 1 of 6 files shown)


ft/ft.h
```diff
@@ -181,11 +181,11 @@ void tokuft_update_product_name_strings(void);
...
-    char db_version[sizeof(toku_product_name) + sizeof("1.2.3 build ") + 256];
-    char environmentdictionary[sizeof(toku_product_name) + sizeof(".environment")];
-    char fileopsdirectory[sizeof(toku_product_name) + sizeof(".directory")];
-    char single_process_lock[sizeof(toku_product_name) + sizeof("___lock_dont_delete_me")];
-    char rollback_cachefile[sizeof(toku_product_name) + sizeof(".rollback")];
+    char db_version[sizeof(toku_product_name) + sizeof("1.2.3 build ") + 256 + 1];
+    char environmentdictionary[sizeof(toku_product_name) + sizeof(".environment") + 1];
+    char fileopsdirectory[sizeof(toku_product_name) + sizeof(".directory") + 1];
+    char single_process_lock[sizeof(toku_product_name) + sizeof("___lock_dont_delete_me") + 1];
+    char rollback_cachefile[sizeof(toku_product_name) + sizeof(".rollback") + 1];
```
georgelorchpercona: Unrelated change?  I'm fine with it but just curious.

dutow: For some, it's not used, but for a few I changed the filename by appending "2" to it - the `+1` ensures that we have enough buffer space.

PR #431, opened 2019-09-13
PS-5932: disable cleaner thread and assert on dirty pages

dutow:


prohaska7:
There are data races on the 'writing_rollback' variable.  The value of this variable after a series of increments and decrements is indeterminate since these operations are not atomic operations.  Helgrind identifies these data races when running the checkpoint_fairness test with helgrind.

PR #434, opened 2019-11-28
PS-5674: Fixed compilation error caused by #warning in sys/sysctrl.h

kamil-holubicki:
#warning directive was added to sys/sysctl.h in glibc>=2.30.
It is enough to disable it for PerconaFT/portability.
third_party/xz references above header as well, but in unused xz (not compiled) part (we compile only liblzma)

### DIFF
--- a/portability/portability.cc
+++ b/portability/portability.cc
@@ -60,7 +60,7 @@ Copyright (c) 2006, 2015, Percona and/or its affiliates. All rights reserved.
 #if defined(HAVE_SYS_SYSCALL_H)
 # include <sys/syscall.h>
 #endif
-#if defined(HAVE_SYS_SYSCTL_H)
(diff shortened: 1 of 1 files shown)


percona-ysorokin:
@dutow 
It looks like `sys/sysctl.h` is needed for only for single `sysctlbyname()` call inside `toku_os_get_phys_memory_size()`.
Personally, I would go with conditionally including it only when `_SC_PHYS_PAGES` is not available. Something like
```
#if defined(HAVE_SYS_SYSCTL_H) && !defined(_SC_PHYS_PAGES)
  #include <sys/sysctl.h>
#endif
```

dutow:
You are right, there's also `toku_get_processor_frequency_sysctl`, but for some reason that calls the sysctl command directly, not the sysctl function in this header.

kamil-holubicki:
done

PR #435, opened 2019-12-12
PS-6142: Fix failing PerconaFT tests

dutow:


prohaska7:
A simpler solution is to move the definition of the force_recovery variable to the FT layer (say ft/ft-ops.cc) rather than duplicating the definition all over the place.

inikep:
Already fixed with https://github.com/percona/PerconaFT/pull/439/files

PR #436, opened 2020-03-12
Fix PS-4328, a crash that occurs when out of locks

prohaska7:
PS-4328 describes a crash in the FT locktree that is caused when the memory
used for locks exceeds the locktree limit AND pending lock requests are
retried.  The lock request retry logic did not handle this case and crashed.

This pull request:
1. Contains locktree test cases that expose this problem.
2. Contains ydb test cases that expose this problem. These test cases shows
that DB->put can return the TOKUDB_OUT_OF_LOCKS error when the locktree
memory exceeds its limit.  This is a normal case.
3. Contains a bug fix in the lock request retry logic that correctly handles
the TOKUDB_OUT_OF_LOCKS error by terminating the pending lock request with
the TOKUDB_OUT_OF_LOCKS error.  The application logic using the FT library
SHOULD handle this error.

Tools:
Ubuntu 19.10, Clang 7,8,9, Gcc 7,8,9, Cmake 3.14

Copyright (c) 2020, Rik Prohaska
All rights reserved.


prohaska7:
Replaced by https://github.com/percona/PerconaFT/pull/441

PR #448, opened 2020-06-02
PS-6789 : Merge MySQL 8.0.20

georgelorchpercona:
- Added new archive dependency for tokudb_dump tool required as of 8.0.20


georgelorchpercona:
Hey @percona-ysorokin Here is the combined jenkins job of this change within the 8.0.20 branch and with the TokuDB (and MyRocks, but irrelevant to this) that gets things building and passing most mtr https://ps80.cd.percona.com/view/8.0/job/percona-server-8.0-param/593/

PR #453, opened 2021-04-26
PS-7621 - Improve detection of -DENABLED_DEBUG_SYNC in CMakeLists

ldonoso:
MySql changed the way of adding the flag `-DENABLED_DEBUG_SYNC`,
which caused the variable not being recognized in CMakeLists.

https://github.com/mysql/mysql-server/blob/a9b0c712de3509d8d08d3ba385d41a4df6348775/CMakeLists.txt#L475

The proposed change is back compatible with the previous code.


percona-ysorokin (approved):
LGTM

PR #461, opened 2021-11-25
Add support for s390x and risc-v

XieJiSS:
Ported from (see-also) https://github.com/facebook/rocksdb/pull/9215


XieJiSS:
Testing method is available at https://github.com/facebook/rocksdb/pull/9215#issuecomment-980511688

PR #464, opened 2021-12-22
PS-8014 fix: Cannot find pthread_yield or pthread_yield_np when compiling PerconaFT on Impish

percona-ysorokin:
https://jira.percona.com/browse/PS-8014

On Ubuntu 21.10 Impish both 'pthread_yield()' and 'pthread_yield_np()'
functions are considered deprecated and trigger warnings when they
are used. This does not allow 'check_function_exists()' CMake macro
detect if these functions are present in the system.

To fix this, we now also check for the 'sched_yield()' function
and fall back to it if the previous two are not found or considered
deprecated.

### DIFF
--- a/portability/toku_pthread.cc
+++ b/portability/toku_pthread.cc
@@ -67,7 +67,9 @@ int toku_pthread_yield(void) {
 #elif defined(HAVE_PTHREAD_YIELD_NP)
     pthread_yield_np();
     return 0;
+#elif defined(HAVE_SCHED_YIELD)
+    return sched_yield();
 #else
-# error "cannot find pthread_yield or pthread_yield_np"
+# error "cannot find pthread_yield or pthread_yield_np or sched_yield"
 #endif
 }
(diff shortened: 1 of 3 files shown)


portability/toku_pthread.cc
```diff
@@ -67,7 +67,9 @@ int toku_pthread_yield(void) {
 #elif defined(HAVE_PTHREAD_YIELD_NP)
     pthread_yield_np();
     return 0;
+#elif defined(HAVE_SCHED_YIELD)
+    return sched_yield();
 #else
-# error "cannot find pthread_yield or pthread_yield_np"
+# error "cannot find pthread_yield or pthread_yield_np or sched_yield"
```
ldonoso, line 73: Please, maybe we should consider `sched_yield` as the first option as the others are deprecated?

ldonoso: Well, it doesn't really matter as this storage is going to be discontinued

ldonoso (approved):

portability/toku_pthread.cc
percona-ysorokin: @ldonoso I just wanted to be less intrusive, so that we won't change anything for existing platforms where `pthread_yield()` / `pthread_yield_np()` are not considered deprecated.

